PR #2, opened 2019-10-22
Removes dependency on crypto library

jaredjensen:
- Replaces `crypto` lib with native `window.crypto`
- Exposes PKCE-related features as new `PKCE` module


src/util/pkce.ts
```diff
@@ -1,28 +1,55 @@
...
-    .replace(/=/g, '');
-}
+  /**
+   * Creates a base-64 encoded url-friendly version of the specified array.
+   *
+   * @param array The array of numbers to encode
+   */
+  public static base64UrlEncode(array: Uint8Array): string {
+    const numbers = Array.prototype.slice.call(array);
+    const encoded = btoa(String.fromCharCode.apply(null, numbers))
```
cerebrl: Just for consistency with window object methods, would you add `window.` to this `btoa` call?

jaredjensen: Done.  Also added basic unit tests and supporting jest config.

src/util/pkce/__tests__/index.ts
```diff
@@ -0,0 +1,37 @@
+import crypto from 'crypto';
+import { TextEncoder } from 'util';
+import PKCE from '..';
+
+Object.defineProperty(window, 'crypto', {
```
cerebrl: Is `crypto` still not available in JSDOM? I noticed you added `jsdom` as an environment for Jest.

cerebrl: After seeing that [this Github issue is still open](https://github.com/jsdom/jsdom/issues/1612), I'm assuming `crypto` has still not been implemented, so I see why this was done.

PR #8, opened 2019-10-28
Prepares 0.9.1 release

jaredjensen:
- Merges PRs for testing and FRPolicy
- Exports `SessionManager`
- Adds `CHANGELOG.md`


src/fr-policy/index.ts
```diff
@@ -90,11 +90,9 @@ abstract class FRPolicy {
   ): string {
     const policyRequirement = policy.policyRequirement;
     const params = policy.params ? { ...policy.params, policyRequirement } : { policyRequirement };
-    const message = (
-      messageCreator[policyRequirement] ||
+    const message = (messageCreator[policyRequirement] ||
```
cerebrl: This is a bit hard to grok the way it's structured. It took me a while to realize the result of this series of ORs results in a function that is then called. This is totally a style thing, but the following would be been more readable to me:

```
const message = (
        messageCreator[policyRequirement] ||
        policyMessage[policyRequirement] ||
        policyMessage[PolicyKey.UnknownPolicy]
    )(property, params);
```

Just a suggestion.

jaredjensen: Agreed it could be easier to grok.  It looks like this may be a case of Prettier doing not-optimal formatting, though.  Let's address in a future release.

PR #13, opened 2019-11-01
Added WebAuthN methods to SDK

chuckmasucci:
I rebased Chuck's initial PR onto my branch for UI-related work, did some webauthn refactoring, and generally got the SDK ready for GA.


src/browser/index.ts
```diff
@@ -0,0 +1,156 @@
+import { CallbackType } from '../auth/enums';
+import { parseRelyingPartyId, getClientDataJson } from '../util/webauthn';
+import FRStep from 'fr-auth/fr-step';
+import { WebAuthnAuthenticationMetadata, WebAuthnRegistrationMetadata } from 'browser/interfaces';
+
+abstract class WebAuthn {
+  public static async register(step: FRStep) {
+    const { hiddenCallback, metadataCallback } = WebAuthn.getCallbacks(step);
```
jaredjensen: I think you can change all the `WebAuthn.foo` references to `this.foo`.

src/browser/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    return step;
+  }
+
+  public static getCallbacks(step: FRStep) {
+    const hiddenCallback = step.getCallbackOfType(CallbackType.HiddenValueCallback);
+    const metadataCallback = step.getCallbackOfType(CallbackType.MetadataCallback);
+    return { hiddenCallback, metadataCallback };
+  }
+
+  public static async getAuthenticationCredentials(publicKey: PublicKeyCredentialRequestOptions) {
```
jaredjensen: For the auth trees to work correctly, we need to handle more possible outcomes.  The `outcome` value maps to connection points on the auth tree node and determines branching logic.
```
    try {
      if (!window.PublicKeyCredential) {
        return 'unsupported';
      } else {
        // do the needful
        return `${clientDataJSON}::${attestationObject}::${credential.id}`;
      }
    } catch (err) {
      if (err.name === 'NotAllowedError') {
        return 'ERROR::NotAllowedError:The operation either timed out or was not allowed';
      }
      return `ERROR::${err}`;
    }
```

src/event/helpers.ts
```diff
@@ -0,0 +1,31 @@
+import { CallbackContainer, Listener } from './interfaces';
+
+/** @hidden */
```
cerebrl: What does `@hidden` mean?

src/util/webauthn.ts
```diff
@@ -0,0 +1,23 @@
...
+// TODO: Test webauthn and find out what type def for Credential is correct
+function getClientDataJson(credential: Credential): string {
+  // https://goo.gl/yabPex - To future-proof, we'll pass along whatever the browser
+  // gives us and let AM disregard randomly-injected properties
+  const uint8Array = new Uint8Array(credential.response.clientDataJSON);
+  const json = String.fromCharCode.apply(null, Array.from(uint8Array));
+  return json;
+}
+
+// AM is currently serializing RP as one of the following formats, depending on
```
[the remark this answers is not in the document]
jaredjensen: For context, the bulk of comments are for typedocs only.  I'm personally not a fan of having comments at all unless they're needed to describe goofy, but required, functionality.  I 100% agree about consistency, but am hesitant to make any changes here until the team creates a coding style guide.

src/fr-webauthn/helpers.ts
```diff
@@ -0,0 +1,47 @@
+import { ParsedCredential } from './interfaces';
+
+function parsePubKeyArray(s: string | unknown[]): PublicKeyCredentialParameters[] | undefined {
```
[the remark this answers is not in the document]
jaredjensen: I think it's consistent with something like `parseInt`, no?  I'm actually a fan of one-letter variables in small pure functions like this because it helps avoid line-wrapping, and the functions are so small you're not at risk of forgetting what `s` is.  Another great topic for the style guide!

src/event/helpers.ts
jaredjensen: It means don't render typedocs for it.  Not sure what the difference is between `@hidden` and `@ignore`.
https://typedoc.org/guides/doccomments/#hidden-and-ignore

src/fr-webauthn/index.ts
```diff
@@ -0,0 +1,309 @@
...
+    metadata: WebAuthnAuthenticationMetadata,
+  ): PublicKeyCredentialRequestOptions {
+    const { acceptableCredentials, challenge, relyingPartyId, timeout } = metadata;
+    const rpId = parseRelyingPartyId(relyingPartyId);
+
+    return {
+      allowCredentials: parseCredentials(acceptableCredentials),
+      challenge: Uint8Array.from(atob(challenge), (c) => c.charCodeAt(0)).buffer,
+      timeout,
+      ...(rpId && { rpId }),
```
cerebrl: This is quite clever. Though, I'm not a huge fan of using clever code as it sets a dangerous precedent. I'm guessing you wanted a super succinct way to only include the property of `rpId` if it exists with as few variable declarations as possible?

src/shared/interfaces.ts
```diff
@@ -1,7 +1,11 @@
+interface AnyObject {
```
cerebrl: Not sure I'm a fan of this. The type itself is actually misleading, unless I'm misunderstanding something. Doesn't this define a type that is an object that can only have a key of string with a value of an empty object, `null` or `undefined`, which is essentially a data-less object?

src/oauth2-client/index.ts
```diff
@@ -20,7 +26,7 @@ abstract class OAuth2Client {
     const { serverConfig, clientId, redirectUri, scope } = Config.get(options);
 
     /* eslint-disable @typescript-eslint/camelcase */
-    const requestParams: any = {
+    const requestParams: AnyObject = {
```
cerebrl: I don't believe this aligns with the definition of `AnyObject` as these properties have actual values, yeah?

src/fr-webauthn/helpers.ts
```diff
@@ -0,0 +1,72 @@
...
+        };
+      });
+    return creds;
+  } catch (error) {
+    throw new Error('Failed to parse credentials');
+  }
+}
+
+function getClientDataJson(credential: ResponseCredential): string {
+  // https://goo.gl/yabPex - To future-proof, we'll pass along whatever the browser
```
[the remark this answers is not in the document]
jaredjensen: I personally don't mind the inconsistency here.  Using JSDoc style literally doubles the number of lines required for a 2-line comment.  If it's agreed that non-documentation comments should be rare and brief, then I'd favor this more terse format.

src/fr-webauthn/index.ts
jaredjensen: It's more about reducing overly-explicit conditionals so the function is easier to grok.  I wouldn't refer to this as "clever" because this style is used all the time in React JSX, and it's pretty easy to understand.  Another thing for the style guide.

PR #40, opened 2020-04-08
Feature (SDKS-312): Device Profile collection callback support

cerebrl:
Summary:

This PR provides the support of a new callback called `DeviceProfileCallback`, a callback for a new authentication tree node for using user device profiles as an additional layer of authentication.

Notes:

- E2E tests now have separate directories with their own `autoscripts` and `index.html` for each feature
- This profile collection feature has a single, instantiable class with both optional configuration and methods that can be overwritten to meet the product requirements

Todos:

- [x] Add profile identification support: generating an ID and save to localStorage
- [x] Add more unit tests for varying configurations
- [x] Add negative test cases for both unit and E2E
- [x] Research for a way to automate geolocation test
- [x] Tested in IE11


src/fr-auth/callbacks/device-profile-callback.ts
```diff
@@ -0,0 +1,45 @@
+import FRCallback from '.';
+import { Callback } from '../../auth/interfaces';
+import { DeviceProfileData } from '../../fr-device/interfaces';
+
+/**
+ * Represents a callback used to collect device profile data.
+ */
+class DeviceProfileCallback extends FRCallback {
```
cerebrl, line 8: New callback for Device Profile.

src/fr-device/defaults.ts
```diff
@@ -0,0 +1,64 @@
+export const browserProps = [
```
cerebrl, line 1: Default configurations for profile properties to collect.

src/fr-device/index.ts
```diff
@@ -0,0 +1,209 @@
...
+  CollectParameters,
+  DeviceProfileData,
+  Geolocation,
+  ProfileConfigOptions,
+} from './interfaces';
+
+/**
+ * @class FRDevice - Collects user device metadata
+ */
+class FRDevice {
```
cerebrl, line 21: Profile collection class for generating the Device Profile. This is the primary code for this feature.

src/fr-device/sample-profile.json
```diff
@@ -0,0 +1,43 @@
+{
```
cerebrl, line 1: Sample, generated profile from Edge on Mac. This will be removed once feature is finalized.

tests/e2e/app/oauth-login/autoscript.js
```diff
@@ -0,0 +1,115 @@
...
+
+  const delay = 0;
+
+  const url = new URL(window.location.href);
+  const amUrl = url.searchParams.get('amUrl');
+  const clientId = url.searchParams.get('clientId') || 'AccountHolderOAuth2';
+  const realmPath = url.searchParams.get('realmPath') || 'root';
+  const resourceUrl =
+    url.searchParams.get('resourceUrl') || 'https://bank.example.com:3001/account';
+  const scope = url.searchParams.get('scope') || 'openid+profile+me.read';
```
jaredjensen, line 14: Same comment about decoded values as above.

src/fr-device/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    if (!navigator) {
+      console.warn('Cannot collect browser metadata. Navigator is not defined.');
+      return {};
+    }
+    return this.config.browserProps.reduce((prev, curr) => {
+      // eslint-disable-next-line
+      // @ts-ignore
+      prev[curr] = navigator[curr] != undefined ? navigator[curr] : null;
+      return prev;
+    }, {});
```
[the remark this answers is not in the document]
cerebrl: Absolutely! I'll work on extracting this out next, now that I've confirmed this approach.

cerebrl:
All changes, with the exception of the comment about function reuse, are done. Thanks for the review, @jaredjensen!

cerebrl:
@jaredjensen PR is ready for final review.

src/fr-device/collector.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @class Collector - base class for FRDevice
+ * Generic collector functions for collecting a device profile attributes
+ */
+class Collector {
+  /**
+   * @method reduceToObject - goes one to two levels into source to collect attribute
+   * @param props - array of strings; can use dot notation for two level lookup
+   * @param src - source of attributes to check
+   */
+  reduceToObject(props: string[], src: any): { [key: string]: string } {
```
chuckmasucci, line 11: Is there no way to define `src`? It seems to be used as an array or an object in bracket notation.

src/fr-device/device-profile.mock.data.ts
```diff
@@ -0,0 +1,119 @@
...
+      },
+      cpuClass: null,
+      deviceMemory: null,
+      hardwareConcurrency: 16,
+      maxTouchPoints: null,
+      oscpu: null,
+    },
+    browser: {
+      appName: 'Netscape',
+      // eslint-disable-next-line
```
chuckmasucci, line 19: Are these comments added to disable the linter from reporting the object not in alphabetical order?

src/fr-device/index.ts
```diff
@@ -0,0 +1,241 @@
...
+      return '';
+    }
+    const context = canvas.getContext && canvas.getContext('2d');
+
+    if (!context) {
+      console.warn('Cannot collect font data. Browser does not support 2d canvas context');
+      return '';
+    }
+    const text = 'abcdefghi0123456789';
+    context.font = '72px Comic Sans';
```
chuckmasucci, line 151: 🤭

cerebrl: Ha ha ha, I actually discovered a good use of Comic Sans :D 

Ref: https://remysharp.com/2008/07/08/how-to-detect-if-a-font-is-installed-only-using-javascript/

src/fr-device/collector.ts
cerebrl: `src` is always an object, but its property values can vary wildly. The best I probably can do is this:

```
{
  [key: string]: any;
}
```

But, that's not much better.

src/fr-device/device-profile.mock.data.ts
cerebrl: Hmm, good question. I don't remember, to be honest. I'll have to remove them and see what's reported. These are literally just copied from browsers for mocking, so I didn't spend much time on them.

src/fr-auth/callbacks/device-profile-callback.ts
```diff
@@ -0,0 +1,45 @@
...
+   * Gets the callback's data.
+   */
+  public getMessage(): string {
+    return this.getOutputByName<string>('message', '');
+  }
+
+  /**
+   * Does callback require metadata?
+   */
+  public isMetadataRequired(): boolean {
```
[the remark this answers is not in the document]
jaredjensen: The `get/set` prefixing is really a Java thing.  IMO, the `is` prefix is superior for signaling intent that this returns a boolean. `getMetadataRequired` sounds like it would determine required metadata and return it.

src/fr-device/collector.ts
```diff
@@ -0,0 +1,37 @@
...
+    return props.reduce((prev, curr) => {
+      if (curr.includes('.')) {
+        const propArr = curr.split('.');
+        const prop = src[propArr[0]] && src[propArr[0]][propArr[1]];
```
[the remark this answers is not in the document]
jaredjensen: The shared interface `StringDict<string>` might be preferable to `object` to replace `{ [key: string]: string }`

src/fr-auth/callbacks/device-profile-callback.ts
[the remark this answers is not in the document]
cerebrl: Yeah, the sole purpose of this is the resulting `boolean`, which is separate from the actual "metadata" itself. So, I feel the `is` is more indicative of that intent.

src/fr-auth/callbacks/factory.ts
```diff
@@ -29,6 +30,8 @@ function createCallback(callback: Callback): FRCallback {
       return new ChoiceCallback(callback);
     case CallbackType.ConfirmationCallback:
       return new ConfirmationCallback(callback);
+    case CallbackType.DeviceProfileCallback:
```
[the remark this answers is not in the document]
cerebrl: I've seen the string values frequently in uppercase, but the keys I've just seen in PascalCase. Here's TypeScript's official enum docs: https://www.typescriptlang.org/docs/handbook/enums.html. Is there a resource you could point me to for this topic?

src/fr-device/collector.ts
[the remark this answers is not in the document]
cerebrl: Good idea.

src/fr-device/collector.ts
```diff
@@ -0,0 +1,37 @@
...
+        prev[propArr[1]] = prop != undefined ? prop : '';
+      } else {
+        prev[curr] = src[curr] != undefined ? src[curr] : null;
```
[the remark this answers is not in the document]
cerebrl: Unfortunately no, and that's because the condition needs to be `false` _only_ in the case of a value being `undefined` or `null`, which `x != undefined` magically does for us. It's important that other falsy values like `''`, `false`, `0` ... result in `true`. Maybe I'll add a comment here explaining that.

src/fr-device/index.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  getDeviceName(): string {
+    if (!navigator) {
+      console.warn('Cannot collect device name. navigator is not defined.');
+      return '';
+    }
+    const userAgent = navigator.userAgent;
+    const platform = navigator.platform;
+
+    switch (true) {
```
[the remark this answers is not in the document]
cerebrl: It's just a way of doing a `switch` on the result of a conditional, rather than directly comparing values in the expression. Though, I am wondering why I didn't use the `string.includes()` syntax below.

src/fr-device/index.ts
```diff
@@ -0,0 +1,241 @@
...
+      console.warn('Cannot generate profile ID. Crypto and/or getRandomValues is not supported.');
+      return '';
+    }
+    if (!localStorage) {
+      console.warn('Cannot store profile ID. localStorage is not supported.');
+      return '';
+    }
+    let id = localStorage.getItem('profile-id');
+    if (!id) {
+      id = window.crypto.getRandomValues(new Uint32Array(3)).join('-');
```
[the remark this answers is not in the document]
cerebrl: Ha ha, it truly is an arbitrary number and really could be anything. I can leave a comment.

src/fr-device/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    if (!context) {
+      console.warn('Cannot collect font data. Browser does not support 2d canvas context');
+      return '';
+    }
+    const text = 'abcdefghi0123456789';
+    context.font = '72px Comic Sans';
+    const baseWidth = context.measureText(text).width;
+
+    const installedFonts = this.config.fontNames.reduce((prev, curr) => {
+      context.font = '72px ' + curr + ', Comic Sans';
```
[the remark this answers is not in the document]
cerebrl: Could catch!

PR #41, opened 2020-04-16
SDK-263 Unify README files in ForgeRock SDK projects.

forgerock-chris:


README.md
```diff
@@ -1,208 +1,143 @@
...
+    <a href="./blob/master/CHANGELOG.md">Change Log</a>
+    ·
+    <a href="#support">Support</a>
+    ·
+    <a href="#documentation" target="_blank">Docs</a>
+  </p>
+  <hr/>
+</p>
+
+The ForgeRock JavaScript SDK enables you to quickly integrate the [ForgeRock Identity Platform](https://www.forgerock.com/digital-identity-and-access-management-platform) into your JavaScript apps.
```
cerebrl, line 19: Should we say "client-side JavaScript apps"? That may seem silly, but it is a distinction that makes a difference.

README.md
```diff
@@ -1,208 +1,143 @@
...
+* ForgeRock Identity Platform
+    * Access Management (AM) 6.5.2+
+
+* Browsers:
+    * Chrome 42
+    * Firefox 39
+    * Safari 10.1
+    * Edge 15
+
+> **Tip**: Older browsers can be supported by using polyfills.
```
cerebrl, line 37: I'm working on a document now that goes into detail regarding this that I believe we should move to official SDK docs sometime soon.

README.md
```diff
@@ -1,208 +1,143 @@
...
+    ```
+    git clone https://github.com/ForgeRock/forgerock-javascript-sdk.git
+    ```
 
-# Generate CA and self-signed certificate
-# (Pick any passphrase and use it whenever prompted)
-npm run certs:make
+3. Open `samples/js/config.js` and edit the values to match your AM instance.
+4. Serve the `samples` directory by using a simple HTTP server, for example:
+   `npx http-server samples -c1 -p 3000`
```
cerebrl, line 64: I should check this, but do we not need a secure connection (HTTPS) for this to work? Is running a local file considered secure?

README.md
forgerock-chris: Doesn't hurt to be specific; I've added client-side.

README.md
forgerock-chris: Great! Drop me a Jira (or reassign your's) when it's ready and I'll put it on the site somewhere.

README.md
forgerock-chris: I'll check again locally, but I think localhost is considered one of the "secure origins" that Chromium is okay with. I'll clarify this in the text, and that it's for testing/eval purposes only.

forgerock-chris:
I don't have write access to the source repos. 
If the changes look good, could someone please merge for me? Cheers.

PR #53, opened 2020-05-22
Feature (request): add middleware to modify request object

cerebrl:
Summary:

As an app developer I want an SDK that is flexible enough, so I can pass additional parameters, headers and other request options to the trees.

Details:

Using the "middleware" pattern commonly used with Express.js and the "reducer" pattern commonly. used with Redux, we are providing a way to intercept a request just before it is sent to the server and run a series of functions, each provided the request object, and action object and a next method.

Todos:

- [x] More tests :)

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -4,6 +4,7 @@ import { StringDict } from '../shared/interfaces';
 import { withTimeout } from '../util/timeout';
 import { getEndpointPath, resolve, stringify } from '../util/url';
 import { Step, StepOptions } from './interfaces';
+import middlewareWrapper from '../util/middleware';
 
 /**
  * Provides direct access to the OpenAM authentication tree API.
@@ -20,8 +21,14 @@ abstract class Auth {
     const { realmPath, serverConfig, tree } = Config.get(options);
     const query = options ? options.query : {};
     const url = this.constructUrl(serverConfig, realmPath, tree, query);
-    const init = this.configureRequest(previousStep);
-    const res = await withTimeout(fetch(url, init), serverConfig.timeout);
+    const req = middlewareWrapper(
+      {
+        url: new URL(url),
+        init: this.configureRequest(previousStep),
+      },
+      previousStep ? 'AUTHENTICATE' : 'START_AUTHENTICATE',
+    );
+    const res = await withTimeout(fetch(req.url.toString(), req.init), serverConfig.timeout);
     const json = await this.getResponseJson<Step>(res);
     return json;
   }
--- a/src/config/interfaces.ts
+++ b/src/config/interfaces.ts
@@ -1,12 +1,18 @@
 import { FRCallbackFactory } from '../fr-auth/callbacks/factory';
 import { Tokens } from '../shared/interfaces';
 
+interface Action {
+  type: string;
+  payload: any;
+}
+
 /**
  * Configuration options.
  */
 interface ConfigOptions {
   callbackFactory?: FRCallbackFactory;
   clientId?: string;
+  middleware?: RequestMiddleware[];
   realmPath?: string;
   redirectUri?: string;
   scope?: string;
@@ -28,6 +34,13 @@ interface CustomPathConfig {
   userInfo?: string;
   revoke?: string;
   sessions?: string;
+}
+
+type RequestMiddleware = (req: RequestObj, action: Action, next: () => RequestObj) => RequestObj;
+
+interface RequestObj {
+  url: URL;
+  init: RequestInit;
 }
 
 /**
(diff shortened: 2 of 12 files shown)


src/auth/index.ts
```diff
@@ -20,7 +21,8 @@ abstract class Auth {
     const { realmPath, serverConfig, tree } = Config.get(options);
     const query = options ? options.query : {};
     const url = this.constructUrl(serverConfig, realmPath, tree, query);
-    const init = this.configureRequest(previousStep);
+    let init = this.configureRequest(previousStep);
+    init = middlewareWrapper(init, previousStep ? 'AUTHENTICATE' : 'START_AUTHENTICATE');
```
adam-cyclones, line 25: could this be a single line const?

src/config/interfaces.ts
```diff
@@ -1,12 +1,18 @@
 import { FRCallbackFactory } from '../fr-auth/callbacks/factory';
 import { Tokens } from '../shared/interfaces';
 
+interface Action {
+  type: string;
+  payload: any;
```
adam-cyclones, line 6: If payload is an object then object type may work better here

src/oauth2-client/index.ts
```diff
@@ -200,6 +201,23 @@ abstract class OAuth2Client {
   ): Promise<Response> {
     const { serverConfig } = Config.get(options);
     const url = this.getUrl(endpoint, query, options);
+    let actionType: string;
```
adam-cyclones, line 204: It might be nice to use this pattern instead.

``` typescript
enum ActionType {
    'AUTHORIZE',
    'LOGOUT',
    'REVOKE_TOKEN',
     'USER_INFO'
}

type Actions = 'accessToken' | 'endSession' | 'revoke';

const actionType = (endpoint: Actions): keyof ActionType => {
    switch (endpoint) {
      case 'accessToken':
        return 'AUTHORIZE';
      case 'endSession':
        return 'LOGOUT';
      case 'revoke':
        return 'REVOKE_TOKEN';
      default:
        return 'USER_INFO';
    }
} 

const valueOfActionType = actionType(endpoint);
```


this way you can retain immutability.

src/session-manager/index.ts
```diff
@@ -13,7 +14,7 @@ abstract class SessionManager {
    */
   public static async logout(options?: ConfigOptions): Promise<Response> {
     const { realmPath, serverConfig } = Config.get(options);
-    const init: RequestInit = {
+    let init: RequestInit = {
```
adam-cyclones, line 17: This looks like it also could be a single line const.

tests/e2e/server/routes.auth.mjs
```diff
@@ -20,12 +20,18 @@ export const baz = {
   canWithdraw: false,
 };
 
-export default function(app) {
+export default function (app) {
   app.post(authPaths.authenticate, wait, async (req, res) => {
     if (!req.body.callbacks) {
```
adam-cyclones, line 25: Watch out for this nested if, its getting a bit deep, perhaps it can be flattened somewhat?

src/config/interfaces.ts
cerebrl: Payload could literally be anything: a string, a number, an object, an array ...

tests/e2e/server/routes.auth.mjs
cerebrl: Ha ha, yeah, this whole route is an absolute mess. I need to spend a whole day refactoring it, but since it's just the mock server, other things are getting priority :) Thanks for bringing this up, I'm going to make a new Jira ticket to do this.

src/oauth2-client/index.ts
cerebrl: I like this idea. I'll make this change.

src/session-manager/index.ts
cerebrl: Do you mean just inline this init object into the middlewareWrapper call below?

src/auth/index.ts
cerebrl: Like this:

```js
const init = middlewareWrapper(
  this.configureRequest(previousStep), 
  previousStep ? 'AUTHENTICATE' : 'START_AUTHENTICATE'
);
```

adam-cyclones: yeah just like that :)

src/config/interfaces.ts
adam-cyclones: Okay all good then.

src/oauth2-client/index.ts
adam-cyclones: I do love a switch statement return, its like the Rust lang match statement.

src/session-manager/index.ts
adam-cyclones: yep just like that.

PR #58, opened 2020-06-05
Feature (auth/middleware): handle noSession and add action payload for auth

cerebrl:
Summary:

To handle the parameter of `noSession` in auth trees, we need to detect the completion, successful or not, without the return of an `idToken`. We also need to provide an action payload for app devs to distinguish between auth types and trees.

### DIFF
--- a/src/util/middleware.mock.data.ts
+++ b/src/util/middleware.mock.data.ts
@@ -1,17 +1,26 @@
 import { Action, RequestObj } from '../config/interfaces';
+import { ActionTypes } from '../config/enums';
 
 type NextFn = () => RequestObj;
+
+const a = 'a' as ActionTypes;
+const b = 'b' as ActionTypes;
+const one = '1' as ActionTypes;
+const two = '2' as ActionTypes;
+const add = 'ADD' as ActionTypes;
+const reassignment = 'REASSIGNMENT' as ActionTypes;
+const mutateAction = 'MUTATE-ACTION' as ActionTypes;
 
 export default [
   (req: RequestObj, action: Action, next: NextFn) => {
     switch (action.type) {
-      case 'a':
-      case 'b':
+      case a:
+      case b:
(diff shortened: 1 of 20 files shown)


tests/e2e/app/basic-login/autoscript.js
```diff
@@ -15,6 +15,23 @@
 
   console.log('Configure the SDK');
   forgerock.Config.set({
+    middleware: [
+      (req, action, next) => {
+        switch (action.type) {
+          case 'START_AUTHENTICATE':
+            if (action.payload.type === 'service' && typeof action.payload.tree === 'string') {
```
adam-cyclones, line 22: If this is

tests/e2e/app/basic-login/autoscript.js
```diff
@@ -15,6 +15,23 @@
...
+              console.log('Starting authentication with service');
+            }
+            break;
+          case 'AUTHENTICATE':
+            if (action.payload.type === 'service' && typeof action.payload.tree === 'string') {
```
adam-cyclones, line 27: the same as this, maybe the if should wrap the switch?

src/util/middleware.mock.data.ts
```diff
@@ -1,17 +1,18 @@
...
+import { ActionTypes } from '../config/enums';
 
 type NextFn = () => RequestObj;
 
 export default [
   (req: RequestObj, action: Action, next: NextFn) => {
     switch (action.type) {
-      case 'a':
-      case 'b':
+      case 'a' as ActionTypes:
```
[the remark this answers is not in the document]
cerebrl: Good point, there is a lot of repetition here.

tests/e2e/app/basic-login/autoscript.js
cerebrl: 👍  good point. I get a little lazy in the test files. I'll fix it.

PR #68, opened 2020-07-07
Test (reg/2nd-factor): registration and second factor automation

cerebrl:


tests/e2e/suites/registration.test.ts
```diff
@@ -0,0 +1,31 @@
+import { setupAndGo } from '../utilities/setup-and-go';
+
+describe('Test Basic login flow', () => {
```
spetrov, line 3: Test basic registration flow...

tests/e2e/suites/registration.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  ['chromium', 'webkit', 'firefox'].forEach((browserType) => {
+    it(`should login successfully and then log out with ${browserType}`, async (done) => {
```
spetrov, line 5: ... similar to the above... "should register successfully..."

tests/e2e/suites/registration.test.ts
cerebrl: Fixed.

tests/e2e/suites/registration.test.ts
cerebrl: Fixed.

tests/e2e/suites/authn-second-factor.test.ts
```diff
@@ -0,0 +1,25 @@
+import { setupAndGo } from '../utilities/setup-and-go';
+
+describe('Test Second Factor login flow', () => {
```
cerebrl, line 3: Added a test suite for second-factor.

PR #71, opened 2020-07-16
Fix (webauthn): conditionally set key-value pairs options obj

cerebrl:
**Summary**:

AM conditionally sets a few key-value pairs for WebAuthn credential options object. When parsing the incoming callbacks, the SDK needed to align with conditions.

**Details**:

`relyingPartyId` and `allowCredentials` needed to be conditionally set in options object in order to avoiding DOM errors and continue iterating the AM trees.


cerebrl:
Discusses with @spetrov. Will do a more in-depth review and test later in the sprint after all the changes are introduced.

PR #77, opened 2020-07-27
Feat (webauthn): support usernameless login with username on device

cerebrl:
**Summary**:

WebAuthn and AM support the idea of "usernameless" capability with WebAuthn tech that supports the storing of the username with the authentication device. When logging in, the user can use the device alone without the need for username or password input.

**Details**:

- Ensure the property of `requireResidentKey` is passed to the key creation options on both JSON-based and script-based configuration
- Ensure `userHandle` from the credential outcome is passed to the key getter options on both JSON-based and script-based configuration

**Testing**:

To test this, the easiest way is to reuse the `tests/e2e/app/authn-webauthn` test page and `PasswordlessWebAuthn` (if not, you'll have to build two different trees):

1. Ensure `PasswordlessWebAuthn` tree is built similar to the picture in the `authn-webauthn` directory
2. In this tree, modify _only_ the WebAuthn registration node to store username on device
3. Now, in Chrome on Mac, run the WebAuthn test script, but stop at the device registration prompt
4. Return to the tree in AM , and update the authentication node to use username from device
5. Complete the device registration using the fingerprint reader
6. The registration should be successful
7. It will then logout and re-authenticate, triggering the WebAuthn request with username from the device
8. Use the same method, but you should now have to choose the user account
9. You should successfully authenticate and logout
10. Test script complete

**Screen recording**:

usernameless-login

### DIFF
--- a/src/fr-webauthn/script-parser.test.ts
+++ b/src/fr-webauthn/script-parser.test.ts
@@ -1,8 +1,10 @@
 import { parseWebAuthnAuthenticateText, parseWebAuthnRegisterText } from './script-parser';
 import {
   authenticateInputWithRpidAndAllowCredentials,
+  authenticateInputWithRpidAllowCredentialsAndQuotes,
   authenticateInputWithoutRpidAndAllowCredentials,
   registerInputWithRpid,
+  registerInputWithRpidAndQuotes,
   registerOutputWithRpid,
   registerInputWithoutRpid,
   registerOutputWithoutRpid,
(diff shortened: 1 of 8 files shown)


src/fr-webauthn/script-parser.ts
```diff
@@ -8,10 +8,13 @@ function parseWebAuthnRegisterText(text: string): PublicKeyCredentialCreationOpt
   const attestation = getIndexOne(text.match(/attestation:\s{0,}"(\w+)"/)) as AttestationType;
   // e.g. `timeout: 60000`
   const timeout = Number(getIndexOne(text.match(/timeout:\s{0,}(\d+)/)));
-  // e.g. `"userVerification":"preferred"`
+  // e.g. from 7.0: `"userVerification":"preferred"`
+  // e.g. from 6.5: `userVerification: "preferred"`
   const userVerification = getIndexOne(
-    text.match(/"userVerification":\s{0,}"(\w+)"/),
+    text.match(/userVerification"{0,}:\s{0,}"(\w+)"/),
```
cerebrl, line 14: Added to support a slight variation between 6.5.2 and 7.0.

src/fr-webauthn/script-parser.ts
```diff
@@ -5,38 +5,43 @@ function parseWebAuthnRegisterText(text: string): PublicKeyCredentialCreationOpt
   const txtEncoder = new TextEncoder();
 
   // e.g. `attestation: "none"`
-  const attestation = getIndexOne(text.match(/attestation:\s{0,}"(\w+)"/)) as AttestationType;
+  const attestation = getIndexOne(text.match(/attestation"{0,}:\s{0,}"(\w+)"/)) as AttestationType;
```
cerebrl, line 8: Duplicated the change needed for 6.5.2 to all parsed keys in case they change in future versions.

src/fr-webauthn/script-parser.test.ts
```diff
@@ -18,6 +20,15 @@ describe('Parsing of the WebAuthn script type text', () => {
     expect(obj.rpId).toBe('example.com');
   });
 
+  it('should parse the WebAuthn authenticate block of text with quoted keys', () => {
```
cerebrl, line 23: Unit tests for testing quoted keys in parsed text.

PR #81, opened 2020-08-05
Feat (webauthn): add authenticatorAttachment & fix bug with multiple devices

cerebrl:
**Summary**:

Addressing a few bugs caught with WebAuthn.

### DIFF
--- a/src/fr-webauthn/script-parser.ts
+++ b/src/fr-webauthn/script-parser.ts
@@ -17,6 +17,10 @@ function parseWebAuthnRegisterText(text: string): PublicKeyCredentialCreationOpt
   const requireResidentKey = getIndexOne(
     text.match(/requireResidentKey"{0,}:\s{0,}(\w+)/),
   ) as string;
+  // e.g. `"authenticatorAttachment":"cross-platform"`
+  const authenticatorAttachment = getIndexOne(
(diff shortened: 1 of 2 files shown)


src/fr-webauthn/script-parser.ts
```diff
@@ -103,7 +109,7 @@ function parseWebAuthnAuthenticateText(text: string): PublicKeyCredentialRequest
       // e.g. `{ \"type\": \"public-key\",
       const type = getIndexOne(str.match(/type"{0,}:\s{0,}"([\w-]+)"/)) as 'public-key';
       // e.g. \"id\": new Int8Array([-107, 93, 68, -67, ... -19, 7, 4]).buffer
-      const idArr = ensureArray(text.match(/id"{0,}:\s{0,}new\s{0,}(Uint|Int)8Array\(([^\)]+)/));
+      const idArr = ensureArray(str.match(/id"{0,}:\s{0,}new\s{0,}(Uint|Int)8Array\(([^\)]+)/));
```
spetrov, line 112: Ah, so that was causing DSKS-572!

cerebrl: Yup, on each iteration, we were accidentally grabbing the first `id` from the whole text, rather than from the individual iteration snippets.

PR #92, opened 2020-09-30
Feat (central-login): full implementation of centralized login

cerebrl:
## Summary

This feature supports an Authorization Code Flow that has an external authentication application. The `TokenManager.getTokens` method now has the configuration option of `login: "redirect"`. This instructs the method to do a full redirect in order for the user to authenticate, if not already authenticated. If the user is already authenticated, then the authorization code and token exchange can happen invisibly to the user without full browser redirects.

## Full Flow

1. App dev calls `TokenManager.getTokens()`
2. SDK checks to see if OIDC tokens are present, if not continue below
3. SDK generates the appropriate authorize URL with PKCE code challenge and state
4. SDK checks to see if it (without full redirect) can obtain access tokens, if not, continue below
5. SDK stores all relevant PKCE values into `sessionStorage`
6. SDK generates a full redirect to OP
7. Redirect back to Web app via its callback URL
8. App dev would integrate a route in app layer to handle the redirection back to the Web app
9. App dev would then call `TokenManager.getTokens({ query: { state, code }})`
10. SDK would pull the PKCE values from `sessionStorage`
11. SDK would complete the flow by calling the token exchange endpoint with the code and PKCE verifier
12. Return new OAuth tokens
13. Done

## Details

In addition to the new `login: "redirect"` option, the capability to use `iframe` for legacy browser support or `fetch` for only modern browsers is now supported in this flow. This can be done by setting `support: "modern"` or `support: "legacy"` on the global config. Not specifying this configuration object property is equivalent to declaring "legacy". 

"Legacy" in this case refers to IE11. It's also worth noting that even though IE11 can be polyfilled for the `fetch` method, reading a redirected URL from this polyfilled method does not work, which is essential to this functionality.

If "legacy" is used, make sure to set the `serverConfig.timeout` very low. The reason for this is there is no bullet-proof method to detect events within an iframe at a granular enough level, so a timeout is used to detect an error state.

If "modern" is used, timeout is not used for error detection.

In order to support all the needed functionality, I had to replace `http-server` with `local-web-server` to have better control over CORS settings.

## Sample app

There is a sample app that can be used for full manual testing:

1. Configure the SDK in `samples/central-login/index.html`
2. Run the local server: `npm run start:samples`
3. Go to `<your domain>/central-login/`

## Tests

There's an added integration test and a few end-to-end tests. No unit tests were written as there were not real "units" to test; most of the functionality was just integrating existing logic.

## Code Examples

```javascript
// optional support property for needed browser support
Config.set({
    // ...
    support: 'legacy',
    // ...
});

// Initiate authorization reusing existing method, will handle all work for app dev
// Will result in a full window redirection, if needed
const tokens = TokenManager.getTokens({
    login: 'redirect',
    serverConfig: {
        timeout: 3000, // keep it quick if "legacy" support is needed
    },
});
 
// Return from full redirect from authentication, pass code and state as options
// Will complete the token exchange, store tokens and return to app dev
const tokens = TokenManager.getTokens({
    query: { 
        code: 'lFJQYdoQG1u7nUm8 ... ',
        state: 'MTY2NDkxNTQ2Nde3D ... ' 
    }
});
```


spetrov (approved):
Looks good to me!

PR #93, opened 2020-10-16
Test (e2e): modified live-capable tests to run against real environment

spetrov:
Summary: Modified e2e test cases to be able to run against live environment. 
To finish: 
- [x] Re-add the 'webkit' and 'firefox' browsers in the tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,19 +14,19 @@
       }
     },
     "@babel/core": {
-      "version": "7.11.6",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.11.6.tgz",
-      "integrity": "sha512-Wpcv03AGnmkgm6uS6k8iwhIwTrcP0m17TL1n1sy7qD0qelDu4XNeW0dN0mHfa+Gei211yDaLoEe/VlbXQzM4Bg==",
+      "version": "7.12.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.0.tgz",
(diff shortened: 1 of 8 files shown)


tests/e2e/suites/register-basic.lc.test.ts
```diff
@@ -14,11 +14,13 @@ import { setupAndGo } from '../utilities/setup-and-go';
 describe('Test basic registration flow', () => {
   ['chromium', 'webkit', 'firefox'].forEach((browserType) => {
     const un = v4();
+    const pw = 'ieH034K&-zlwqh3V_';
```
cerebrl, line 17: @spetrov Are you adding this because of that hardcoded check in the mock server?

spetrov: @cerebrl Yes, the mock server is expecting this value. Without adding the value explicitly here, it is taking the password from /tests/e2e/env.config.ts and the test against mock server fails...

cerebrl: Would you mind changing the mock server to take the password from the configuration, rather than having to make the change here? That way, we can keep the configuration for all of this in the same place.

I just pulled and ran your branch locally, and the following change in `routes.auth.mjs` works as expected: https://share.getcloudapp.com/o0umYKZo.

Other than this, the PR looks good.

spetrov: Yes, this is so much better than it was! 👍 
Fixed it as suggested!

cerebrl (approved):

PR #95, opened 2020-11-13
fix (webauthn): step detection, 6.5.3 support, more tests

cerebrl:
**Summary:**

Ensure all configurations and AM versions (that are compatible) work with the SDK, and its APIs without issue. The newly introduced 6.5.3 was causing a few issues due to a slight change in the API format, so that needed to be handled appropriately. There's also the "usernameless" configuration that was breaking how the SDK recognizes the appropriate WebAuthn step.

To ensure these slight variations do not cause regression bugs, we added more robust automation to ensure intended functionality does not break on upcoming version changes.

Details:

- Fixed step detection bug when "usernameless" is used
- Fixed 6.5.3 parsing bug due to a slight variation of API between 6.5.3 and 7.0
- Upgraded ESLint due to some incompatibility issues between TypeScript 4 and ESLint 3
- Updated ESLint comments to reflect new version
- Added lots of automation around parsing in WebAuthn helper functions

**Todos:**

- [x] Rebase with develop
- [x] Add more detail to PR
- [x] Discuss changes with team

**Notes:**

I had to update the `eslint` tooling, which led to lint comment changes in quite a bit of files, unfortunately.


src/fr-webauthn/fr-webauthn.test.ts
```diff
@@ -0,0 +1,96 @@
+import { WebAuthnStepType } from './enums';
```
rodrigoareis, line 1: Should we include copyright header on the test files for JS SDK as well?

cerebrl: Ah, good catch! Will do that now.

spetrov (approved):
Looks good to me!

PR #99, opened 2020-12-14
fix (central-login): handle redirect errors with FF and Webkit

cerebrl:
Summary:

FireFox and Safari have slightly different error messages related to the redirect to the login page and CORS. Since these errors are expected and should result in a redirect, we needed to add a check for them in the flow.


spetrov (approved):
Looks good to me. I tested FF and Safari and redirects within fetch work fine!

PR #100, opened 2020-12-16
Task (central-login): word change from 'native' to 'embedded'

cerebrl:
Summary:

After internal discussion, we decided to change the string value for determining login handled by the app itself as "embedded", rather than "native".

Details:

The literal string value for the interface declaration has been changed, but no code actually references this value directly, so no other change is required.

### DIFF
--- a/src/token-manager/index.ts
+++ b/src/token-manager/index.ts
@@ -18,7 +18,7 @@ import { parseQuery } from '../util/url';
 
 interface GetTokensOptions extends ConfigOptions {
   forceRenew?: boolean;
-  login?: 'native' | 'redirect' | undefined;
+  login?: 'embedded' | 'redirect' | undefined;
   query?: StringDict<string>;
 }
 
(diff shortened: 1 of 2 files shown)


spetrov (approved):
Looks good to me! 👍

src/token-manager/index.ts
```diff
@@ -18,7 +18,7 @@ import { parseQuery } from '../util/url';
 
 interface GetTokensOptions extends ConfigOptions {
   forceRenew?: boolean;
-  login?: 'native' | 'redirect' | undefined;
+  login?: 'embedded' | 'redirect' | undefined;
```
cerebrl, line 21: Only code change required. Since there are only two choices, the generic `else` in this condition is what triggers the embedded functionality: https://github.com/ForgeRock/forgerock-javascript-sdk/pull/100/files#diff-8c14601248b947426b7c11ac7d595b651088ef97e5dc330af90524d20af79b6bR128

PR #103, opened 2021-01-12
[CI/CD] Junit reporter for jest tests

spetrov:
* Added jest-juint reports, so that we can track results in Jenkins builds. 

**Note:** The reports are generated under `reports/${Date.now()}` directory so that we can run and preserve all reports (unit, integration and e2e...). Take a look [here](https://jenkins.petrov.ca/view/JS/job/JS-Core-SDK/40/testReport/) for an example test report after successful build.


cerebrl (approved):
Looks good!

PR #105, opened 2021-02-19
feat: Allow custom tree in centralised UI

dbalmain:
When calling getTokens with login: redirect, this change allows
you to set the tree/journey.

```
const tokens = await forgerock.TokenManager.getTokens({
  forceRenew: true,
  login: 'redirect',
  tree: 'CustomTree', // new
});
```

### DIFF
--- a/src/oauth2-client/index.ts
+++ b/src/oauth2-client/index.ts
@@ -15,6 +15,7 @@ import { StringDict } from '../shared/interfaces';
 import { Noop } from '../shared/types';
 import TokenStorage from '../token-storage';
 import { isOkOr4xx } from '../util/http';
+import middlewareWrapper from '../util/middleware';
 import PKCE from '../util/pkce';
 import { withTimeout } from '../util/timeout';
 import { getEndpointPath, resolve, stringify } from '../util/url';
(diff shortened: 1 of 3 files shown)


cerebrl:
Hello David, thanks for submitting this PR. We are currently using the ACR claims to instruct OAuth clients to use particular trees by passing the key-value pair in the `query` property of the options object:

```
const tokens = await forgerock.TokenManager.getTokens({
  forceRenew: true,
  login: 'redirect',
  query: { acr_values: 'CustomTree' },
});
```

But, this might be more user-friendly, so I'll bring this idea up with my team. For now, I'm going to switch this to a draft PR while we review this feature and implementation.

Thanks again.

cerebrl:
PR for functionality mentioned above: https://github.com/ForgeRock/forgerock-javascript-sdk/pull/106.

cerebrl:
After thinking on this PR, it would conflict with the base configuration of `tree`. Those that use embedded login would get unexpected behavior if their base tree if different from the authentication tree used prior to the `getTokens` call. Since the ACR values enable this functionality without risk of unexpected behavior, we will continue to provide this feature through the ACR query parameter shown above.

But, we do thank you for taking the time to suggest this feature through a PR. As always, reach out to us if you have any questions or concerns. Thanks again.

PR #106, opened 2021-02-22
Feat (acr_values/authorize): pass query properties through to authorize call

cerebrl:
**Summary:**

This allows for the usage of `acr_values` in order to specify a tree when calling the authorize endpoint.

**Details:**

Pass `query` object properties from the  `options` object in `getTokens` through to the authorize call as query parameters. This allows the use of ACR Values to be used to call specific trees.

**Notes:**

There are a lot of test file changes, but it's just the changing of fallback username and password values to align with the test automation's configuration object and its test users.

Relevant file changes to feature:

- https://github.com/ForgeRock/forgerock-javascript-sdk/pull/106/files#diff-92cd59f4227ad81ad8a7bddc61f9b5396662a6720247007d545ee1da7c7e0b1a
- https://github.com/ForgeRock/forgerock-javascript-sdk/pull/106/files#diff-253bb1127ccf96479910d15d3b8c63e6d9152b65ddfa84ec676b9c0f3302d9a8

### DIFF
--- a/tests/e2e/app/authn-central-login/autoscript.js
+++ b/tests/e2e/app/authn-central-login/autoscript.js
@@ -18,11 +18,12 @@
   const url = new URL(window.location.href);
   const amUrl = url.searchParams.get('amUrl') || 'https://auth.example.com:9443/am';
   const code = url.searchParams.get('code') || '';
-  const clientId = url.searchParams.get('clientId') || 'WebOAuthClient';
+  const clientId = url.searchParams.get('clientId') || 'CentralLoginOAuthClient';
   const realmPath = url.searchParams.get('realmPath') || 'root';
   const scope = url.searchParams.get('scope') || 'openid profile me.read';
   const state = url.searchParams.get('state') || '';
   const support = url.searchParams.get('support') || 'legacy';
+  const acr_values = url.searchParams.get('acr') || 'SpecificTree';
 
   console.log('Configure the SDK');
   forgerock.Config.set({
@@ -49,7 +50,7 @@
       .from([1])
       .pipe(
         rxMap(() => {
-          console.log('Set cookie');
+          console.log('Set mock cookie to represent existing session');
           document.cookie = 'iPlanetDirectoryPro=abcd1234; domain=example.com; path=/';
           if (code && state) {
             window.sessionStorage.setItem(
@@ -64,9 +65,9 @@
           console.log('Get OAuth tokens');
           let tokens;
           if (code && state) {
-            tokens = forgerock.TokenManager.getTokens({ query: { code, state } });
+            tokens = forgerock.TokenManager.getTokens({ query: { code, state, acr_values } });
           } else {
-            tokens = forgerock.TokenManager.getTokens();
+            tokens = forgerock.TokenManager.getTokens({ query: { acr_values } });
           }
           return tokens;
         }),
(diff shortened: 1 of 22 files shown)


tests/e2e/server/routes.auth.mjs
```diff
@@ -226,12 +226,17 @@ export default function (app) {
...
   app.get(authPaths.authorize, wait, async (req, res) => {
     if (req.cookies.iPlanetDirectoryPro) {
-      const url = new URL(`${req.query.redirect_uri}`);
-      url.searchParams.set('client_id', 'bar');
-      url.searchParams.set('code', 'foo');
-      url.searchParams.set('iss', `${AM_URL}/oauth2`);
-      url.searchParams.set('state', req.query.state);
-      res.redirect(url);
+      // Is Central Login test
+      if (req.query.acr_values === 'SpecificTree') {
```
spetrov, line 230: This change is breaking the following e2e tests: `authn-oauth`, `config-token-storage` and `config-custom-paths`. They were failing with 400 status response and the error message about the missing arg_values.

spetrov (changes requested):
The changes in the SDK look good to me! 👍 
I tested manually against live environment and was able to trigger "custom" tree by using `arc` claim passed as a query parameter to the /authorize call... 
The /authorize endpoint in the mock server however returns 400 whenever `acr_values` is missing, which breaks some e2e test.

tests/e2e/server/routes.auth.mjs
cerebrl: Ah, nice catch. Sorry for this. I've pushed an update.

tests/e2e/app/authn-central-login/autoscript.js
```diff
@@ -18,15 +18,24 @@
   const url = new URL(window.location.href);
   const amUrl = url.searchParams.get('amUrl') || 'https://auth.example.com:9443/am';
   const code = url.searchParams.get('code') || '';
-  const clientId = url.searchParams.get('clientId') || 'WebOAuthClient';
+  const clientId = url.searchParams.get('clientId') || 'CentralLoginOAuthClient';
```
cerebrl, line 21: This fixes the e2e failures as I can now isolate this test from the other OAuth tests.

tests/e2e/server/routes.auth.mjs
```diff
@@ -225,6 +225,13 @@ export default function (app) {
   });
 
   app.get(authPaths.authorize, wait, async (req, res) => {
+    // Detect if Central Login to enforce ACR value presence
+    if (
+      req.query.client_id === 'CentralLoginOAuthClient' &&
+      req.query.acr_values !== 'SpecificTree'
+    ) {
```
cerebrl, line 232: Here's where I'm isolating just the Central Login for testing ACR values. All other OAuth test automation uses a different client ID.

cerebrl:
@spetrov PR's been updated to address feedback. Thanks!

tests/e2e/app/authn-central-login/autoscript.js
spetrov: 👍 great!

tests/e2e/server/routes.auth.mjs
spetrov: 👍

spetrov (approved):
Looks good to me!

PR #114, opened 2021-04-13
feat (social-login): support Apple Sign In

cerebrl:
**Summary**

This adds the needed support for Apple Sign In and its requirement for using a form POST, rather than a traditional GET redirect. This requires AM to support this form POST, and redirect back to the client web app.

**Details**

- Apple provider configuration requires AM to handle the form POST
- AM will process that form POST and encrypt and encode the data and do a GET redirect back to the client web app with the URL parameter of `form_post_entry`
- The SDK provides an existing `resume` that has been modified to include this `form_post_entry` along with the previously saved step and its `authId`


spetrov (approved):
Looks good to me! 👍

PR #115, opened 2021-04-21
task (webauthn): improve error handling

cerebrl:
**Summary**

Original implementation of WebAuthn just throw errors and left if up to the developer to handle them. This PR adds a more an enhancement to the error handling for the convenience of the developer.

**Details**

All errors have been standardized around the spec found here: https://heycam.github.io/webidl/#idl-DOMException-error-names. I've also tried to stay aligned with iOS and Android's error handling as well.

This implementation catches the error, formats it appropriately for injecting into the `hiddenCallback`, and then throws with that same error object. This is so the developer is made aware of the error, and she can handle it if wanted, but if she just wants to submit it as-is to AM, she can just call `next`.

**Example usage**

```js
try {
    step = await FRWebAuthn.register(step);
} catch (err) {
    // Handle if you want special treatment for errors
    // Ignore if desired
}
// Error has already been injected into the callback from the previous register/authenticate method
// So, you can just ignore the error and continue to next step
step = await FRAuth.next(step);
```

**Screenshots**

SecurityError example:
SecurityError

NotAllowedError example:
NotAllowedError

TypeError example:
TypeError

Unsupported example:
Unsupported

NOTE: There are other error types, but I was unable to reproduce them for examples.

### DIFF
--- a/src/fr-webauthn/index.ts
+++ b/src/fr-webauthn/index.ts
@@ -12,7 +12,7 @@ import { CallbackType } from '../auth/enums';
 import HiddenValueCallback from '../fr-auth/callbacks/hidden-value-callback';
 import MetadataCallback from '../fr-auth/callbacks/metadata-callback';
 import FRStep from '../fr-auth/fr-step';
-import { WebAuthnOutcome, WebAuthnStepType } from './enums';
+import { WebAuthnOutcome, WebAuthnOutcomeType, WebAuthnStepType } from './enums';
 import {
   arrayBufferToString,
   parseCredentials,
@@ -106,13 +106,22 @@ abstract class FRWebAuthn {
         const credential = await this.getAuthenticationCredential(publicKey);
         outcome = this.getAuthenticationOutcome(credential);
       } catch (error) {
-        throw new Error(`WebAuthn: ${error.message}`);
+        // NotSupportedError is a special case
+        if (error.name === WebAuthnOutcomeType.NotSupportedError) {
+          hiddenCallback.setInputValue(WebAuthnOutcome.Unsupported);
+          throw error;
+        }
+        hiddenCallback.setInputValue(`${WebAuthnOutcome.Error}::${error.name}:${error.message}`);
+        throw error;
       }
 
       hiddenCallback.setInputValue(outcome);
       return step;
     } else {
-      throw new Error('WebAuthn: Invalid payload');
+      const e = new Error('Incorrect callbacks for WebAuthn authentication');
+      e.name = WebAuthnOutcomeType.DataError;
+      hiddenCallback?.setInputValue(`${WebAuthnOutcome.Error}::${e.name}:${e.message}`);
+      throw e;
     }
   }
 
@@ -141,13 +150,22 @@ abstract class FRWebAuthn {
         const credential = await this.getRegistrationCredential(publicKey);
         outcome = this.getRegistrationOutcome(credential);
       } catch (error) {
-        throw new Error(`WebAuthn: ${error.message}`);
+        // NotSupportedError is a special case
+        if (error.name === WebAuthnOutcomeType.NotSupportedError) {
+          hiddenCallback.setInputValue(WebAuthnOutcome.Unsupported);
+          throw error;
+        }
+        hiddenCallback.setInputValue(`${WebAuthnOutcome.Error}::${error.name}:${error.message}`);
+        throw error;
       }
 
       hiddenCallback.setInputValue(outcome);
       return step;
     } else {
-      throw new Error('WebAuthn: Invalid payload');
+      const e = new Error('Incorrect callbacks for WebAuthn registration');
+      e.name = WebAuthnOutcomeType.DataError;
+      hiddenCallback?.setInputValue(`${WebAuthnOutcome.Error}::${e.name}:${e.message}`);
+      throw e;
     }
   }
 
@@ -224,6 +242,12 @@ abstract class FRWebAuthn {
   public static async getAuthenticationCredential(
     options: PublicKeyCredentialRequestOptions,
   ): Promise<PublicKeyCredential | null> {
+    // Feature check before we attempt registering a device
+    if (!window.PublicKeyCredential) {
+      const e = new Error('PublicKeyCredential not supported by this browser');
+      e.name = WebAuthnOutcomeType.NotSupportedError;
+      throw e;
+    }
     const credential = await navigator.credentials.get({ publicKey: options });
     return credential as PublicKeyCredential;
   }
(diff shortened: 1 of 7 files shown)


src/fr-webauthn/index.ts
```diff
@@ -106,13 +106,22 @@ abstract class FRWebAuthn {
...
         outcome = this.getAuthenticationOutcome(credential);
       } catch (error) {
-        throw new Error(`WebAuthn: ${error.message}`);
+        // NotSupportedError is a special case
+        if (error.name === WebAuthnOutcomeType.NotSupportedError) {
+          hiddenCallback.setInputValue(WebAuthnOutcome.Unsupported);
+          throw error;
+        }
+        hiddenCallback.setInputValue(`${WebAuthnOutcome.Error}::${error.name}:${error.message}`);
+        throw error;
```
cerebrl, line 115: This is the top level where all the errors get funneled into and processed.

src/fr-webauthn/index.ts
```diff
@@ -141,13 +150,22 @@ abstract class FRWebAuthn {
...
         outcome = this.getRegistrationOutcome(credential);
       } catch (error) {
-        throw new Error(`WebAuthn: ${error.message}`);
+        // NotSupportedError is a special case
+        if (error.name === WebAuthnOutcomeType.NotSupportedError) {
+          hiddenCallback.setInputValue(WebAuthnOutcome.Unsupported);
+          throw error;
+        }
+        hiddenCallback.setInputValue(`${WebAuthnOutcome.Error}::${error.name}:${error.message}`);
+        throw error;
```
cerebrl, line 159: Top level for registration where all errors get funneled and processed.

src/fr-webauthn/index.ts
```diff
@@ -235,15 +259,13 @@ abstract class FRWebAuthn {
    * @return The outcome string
    */
   public static getAuthenticationOutcome(credential: PublicKeyCredential | null): string {
-    if (!window.PublicKeyCredential) {
-      return WebAuthnOutcome.Unsupported;
```
cerebrl, line 239: I found it odd that the feature check was done after the API was accessed. So, I moved it to just before the API was used on line 246 above.

src/fr-webauthn/index.ts
```diff
@@ -235,15 +259,13 @@ abstract class FRWebAuthn {
...
+      const e = new Error('No credential generated from authentication');
+      e.name = WebAuthnOutcomeType.UnknownError;
+      throw e;
     }
 
     try {
-      if (credential === null) {
-        throw new Error('WebAuthn: No credential provided');
-      }
-
```
cerebrl, line 246: Just moved this up a few lines and out of the try-catch.

src/fr-webauthn/index.ts
```diff
@@ -274,6 +298,12 @@ abstract class FRWebAuthn {
   public static async getRegistrationCredential(
     options: PublicKeyCredentialCreationOptions,
   ): Promise<PublicKeyCredential | null> {
+    // Feature check before we attempt registering a device
+    if (!window.PublicKeyCredential) {
+      const e = new Error('PublicKeyCredential not supported by this browser');
+      e.name = WebAuthnOutcomeType.NotSupportedError;
+      throw e;
+    }
```
cerebrl, line 306: Again, moved the feature check to be closer to the API usage, rather than after.

src/fr-webauthn/index.ts
```diff
@@ -235,15 +259,13 @@ abstract class FRWebAuthn {
    * @return The outcome string
    */
   public static getAuthenticationOutcome(credential: PublicKeyCredential | null): string {
-    if (!window.PublicKeyCredential) {
-      return WebAuthnOutcome.Unsupported;
+    if (credential === null) {
+      const e = new Error('No credential generated from authentication');
+      e.name = WebAuthnOutcomeType.UnknownError;
+      throw e;
```
cerebrl, line 265: This shouldn't really happen, but it's just for a possible edge case.

src/fr-webauthn/index.ts
```diff
@@ -346,8 +376,10 @@ abstract class FRWebAuthn {
   ): PublicKeyCredentialCreationOptions {
     const { pubKeyCredParams: pubKeyCredParamsString } = metadata;
     const pubKeyCredParams = parsePubKeyArray(pubKeyCredParamsString);
-    if (!pubKeyCredParams || !pubKeyCredParams.length) {
-      throw new Error('Missing pubKeyCredParams');
+    if (!pubKeyCredParams) {
+      const e = new Error('Missing pubKeyCredParams property from registration options');
+      e.name = WebAuthnOutcomeType.DataError;
+      throw e;
```
cerebrl, line 382: We originally checked for both property existence and array length, but the spec says length can be zero, so I just stuck with property check.

tests/e2e/app/authn-webauthn/autoscript.js
```diff
@@ -61,15 +56,20 @@
...
             throw new Error('WebAuthn step is incorrectly identified');
           }
           console.log('Handle WebAuthn Registration');
-          return forgerock.FRWebAuthn.register(step);
+          try {
+            step = await forgerock.FRWebAuthn.register(step);
+          } catch (err) {
+            console.log(err);
+          }
+          return step;
```
cerebrl, line 72: Adding some try-catches to help observe errors and their submission to AM for testing.

src/fr-webauthn/index.ts
```diff
@@ -141,13 +150,22 @@ abstract class FRWebAuthn {
         const credential = await this.getRegistrationCredential(publicKey);
         outcome = this.getRegistrationOutcome(credential);
       } catch (error) {
-        throw new Error(`WebAuthn: ${error.message}`);
+        // NotSupportedError is a special case
+        if (error.name === WebAuthnOutcomeType.NotSupportedError) {
```
george-bafaloukas-forgerock, line 154: Looks good, just a tiny optional remark.
I would suggest renaming the 'NotSupportedError' to 'Unsupported' to be in sync with the WebAuthnOutcome naming.

george-bafaloukas-forgerock (approved):

dmtrio (approved):
This looks great. I like how nicely all the variations of possible errors are formatted. Again the SDK makes things easier!

spetrov (approved):
Looks good to me! 👍

src/fr-webauthn/index.ts
cerebrl: Hey @george-bafaloukas-forgerock , sorry I missed this. So, `NotSupportedError` is the official error from the spec, but AM is expecting us to map it to the `unsupported` outcome, and not the error outcome. I hope that make sense.

cerebrl:
Thanks @dmtrio, @spetrov and @george-bafaloukas-forgerock for reviewing this. I'll go ahead and merge this in.

PR #123, opened 2021-05-20
task (getTokens): encode data before storage and redirect

cerebrl:
Summary:

Encode data before storing in `sessionStorage` when fully redirecting to `/authorize` URL.

Details:

- base64 encode before storage and after retrieval
- write e2e tests coverage for full redirect


lsanches (approved):
Approved!

PR #124, opened 2021-05-20
task (getTokens): revert encoding on stored data

cerebrl:
**Summary:**

Remove the encoding and decoding of data, but preserve the additional e2e tests from this task.


lsanches (approved):
Approved!

PR #138, opened 2021-10-05
fix (middleware): expand middleware passing to call-site

cerebrl:
**Summary**:

Previously, the middleware for intercepting requests from the SDK was only available for configuration at `Config.set`. This allows the middleware to also be passed at call-sites. Middleware passed at call-sites override the middleware set at config initialization. 

The API or function signature of the non-exported utility function `middlewareWrapper` was changed. Since this is a non-exported member of the top `index.js`, I consider it private so a change to its API is allowed in a non-major release.

Developers now have two options:

```js
// Option 1
Config.set({
  // normal config ...
  middleware: [
    (req, action, next) => {
      switch (action.type) {
        case 'AUTHENTICATE':
          // middleware logic
        default:
          // more logic
      }
      next();
    }
  ]
});
```

```js
// Option 2
// Pass the middleware within the options object of many functions
// e.g. FRAuth.next()
FRAuth.next(
  null,
  {
    // other configuration overrides ...
    middleware: [
      (req, action, next) => {
        switch (action.type) {
          case 'AUTHENTICATE':
            // middleware logic
          default:
            // more logic
        }
        next();
      }
    ]
  }
);
```

Additional Changes:

- Updated middleware tests and mock server to more robustly test this feature
- Updated source code to align with new TypeScript 4.4's stricter typings


src/util/middleware.ts
```diff
@@ -8,32 +8,35 @@
...
+ * @param action - The action object that is passed into the middleware communicating intent
+ * @param action.type - A "Redux" style type that contains the serialized action
+ * @param action.payload - The payload of the action that can contain metadata
  */
-// eslint-disable-next-line
-function middlewareWrapper(request: RequestObj, type: ActionTypes, payload?: any): RequestObj {
-  const { middleware } = Config.get();
-
+function middlewareWrapper(
+  middleware: RequestMiddleware[] | undefined,
```
ryanbas21, line 24: The only thing i'm having a hard time wrapping my head around is why we want the middleware to be the first argument to the middlewareWrapper.

This feels ergonomically weird to me as it can be `undefined`. 

What would the benefits of this signature be to say a signature where we curried a function expecting middleware?

Say:

`middlewareWrapper(request: RequestObj, type: ActionTypes, payload?: any) => (Middleware) => RequestObj;`

We may not need to curry it at all, but the ability to pass undefined as a first argument feels a little weird. But if we have good reason for it i'm ok.

cerebrl:
run tests

PR #140, opened 2021-11-10
Sdks 1347/run playwright tests via GitHub action

ryanbas21:
Adding playwright github actions.

Updated Jest, and had to fix tests where we returned a promise but also used `done` as that is not allowed in jest 27.

Jest 27 allows for the circus test runner which allows for retries which we need in e2e scenarios because of flaky tests.

We want to be able to run our e2e suite in CI so we have a public facing display of our tests, and also a consistent env where we run tests on a given PR.

This PR has a few WIP areas.

First of note, is that Playwright in CI offers the ability to download browsers specifically for each test rather than with the npm install. This is so that CI builds are faster as we only need the browser we are testing in the given run. We will have to supply the given browser_path to runs in CI so that playwright can pickup the correct browser.

This is also going to allow us to target specific browser versions.

Currently, I skipped that for the purposes of this PR as it was causing some problems.

We are testing the latest browsers, on 3 operating systems on latest npm in the given matrix. I think this is enough for now.

Second: I will need to update the jest e2e config in a follow up when we begin nx work. Reason being, that we run with the jest preset of ts-jest but we will need to run the preset of ts-jest with the jest playwright preset as well. This is so we can interop with nx and start caching tests/builds as possible.

https://bugster.forgerock.org/jira/browse/SDKS-1347


src/http-client/helpers.test.ts
```diff
@@ -26,46 +26,44 @@ import {
...
 
     const output = buildAuthzOptions(txnAuthObj, 'https://openam.example.com/am/', 0);
     expect(output.url).toStrictEqual(expectedUrl);
   });
 
   it('build auth by txn req options', () => {
     const txnAuthObj = authzTxnJSON;
-    // eslint-disable-next-line max-len, prettier/prettier
-    const expectedUrl = 'https://openam.example.com/am/json/realms/root/authenticate?authIndexType=composite_advice&authIndexValue=%3CAdvices%3E%3CAttributeValuePair%3E%3CAttribute%20name%3D%22TransactionConditionAdvice%22%2F%3E%3CValue%3Eabc%3C%2FValue%3E%3C%2FAttributeValuePair%3E%3C%2FAdvices%3E';
+    const expectedUrl =
```
ryanbas21, line 39: prettier formatted this.

src/http-client/helpers.test.ts
```diff
@@ -26,46 +26,44 @@ import {
...
+      // eslint-disable-next-line max-len, prettier/prettier
+      'https://openam.example.com/am/json/realms/root/authenticate?authIndexType=composite_advice&authIndexValue=%3CAdvices%3E%3CAttributeValuePair%3E%3CAttribute%20name%3D%22TransactionConditionAdvice%22%2F%3E%3CValue%3Eabc%3C%2FValue%3E%3C%2FAttributeValuePair%3E%3C%2FAdvices%3E';
     const output = buildAuthzOptions(txnAuthObj, 'https://openam.example.com/am/', 0);
     expect(output.url).toStrictEqual(expectedUrl);
   });
 
-  it('examines response for IG auth by tree', async (done) => {
```
ryanbas21, line 44: As an upgrade to Jest 27, it is illegal to use both, done and async (returning a promise). I removed `done` from the tests since we implicitly return a promise with `async` `await`

tests/e2e/server/app.certs.mjs
```diff
@@ -10,9 +10,11 @@
 
 import { readFileSync } from 'fs';
 import { fileURLToPath } from 'url';
+import path from 'path';
```
ryanbas21, line 13: We use the certificates from `lws`

tests/e2e/utilities/setup-and-go.ts
```diff
@@ -34,6 +34,7 @@ export async function setupAndGo(
   },
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): Promise<{ browser: any; messageArray: string[]; networkArray: string[] }> {
+  jest.retryTimes(3);
```
ryanbas21, line 37: E2E's are notoriously flaky. By leveraging the `jest/circus-runner` which is the default test runner in jest 27, we can use `retries` for tests that fail for reasons that are not because of code/env of ours. 

For example, the browser failing to spin up. 3 times should be enough but we can edit it and see how it looks.

.github/workflows/main.yml
```diff
@@ -0,0 +1,94 @@
...
+        node-version: ${{ matrix.nodeversion }}
+    - name: Setup custom host for sdkapp.example.com
+      run: echo "127.0.0.1 sdkapp.example.com sdks.mytestrun.com sdkapp.petrov.ca auth.example.com openig.example.com api.example.com user.example.com rest.example.com bank.example.com openam.example.com spa.example.com sdk.mytestrun.com sdks.mytestrun.com forgerock-sdk.local" | sudo tee -a /etc/hosts
+    - name: Install NPM
+      run: npm i -g npm@8
+    - name: Install Dependencies
+      run: npm ci
+      env:
+        DEBUG: pw:install
+        PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
```
ryanbas21, line 32: We skip browser downloads because in the future, it would be better for the `npx playwright install --with-deps` line to only install the browser from `${{ matrix.browser }}` so we can speed up CI times.

I skipped this in this work here and just installed all the browsers because it was a little bit of a premature optimization but also because I was running into some problems that were not worth debugging to slow this work down. It will be addressed in future work around the pipeline.

.github/workflows/main.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Setup custom host for sdkapp.example.com
+        run: echo "127.0.0.1 sdkapp.example.com sdks.mytestrun.com sdkapp.petrov.ca auth.example.com openig.example.com api.example.com user.example.com rest.example.com bank.example.com openam.example.com spa.example.com sdk.mytestrun.com sdks.mytestrun.com forgerock-sdk.local" | sudo tee -a /etc/hosts      
+      - name: Install NPM
+        run: npm i -g npm@8
+      - name: Install Dependencies
+        run: npm ci
+        env:
+          DEBUG: pw:install
+          # PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
+      # - run: npx playwright install --with-deps ${{ matrix.browser }} chromium
```
ryanbas21, line 64: Because I am not skipping browser downloads here, I don't need to install playwright on the side.

I will align these efforts across all the test scenarios.

.github/workflows/main.yml
```diff
@@ -0,0 +1,94 @@
...
+        browser: [chromiumm, msedge, firefox] # firefox, webkit, msedge]
+        nodeversion: ['16']
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.nodeversion }}
+    - name: Setup custom host for sdkapp.example.com
+      run: echo "`n127.0.0.1  sdkapp.example.com sdks.mytestrun.com sdkapp.petrov.ca auth.example.com openig.example.com api.example.com user.example.com rest.example.com bank.example.com openam.example.com spa.example.com sdk.mytestrun.com sdks.mytestrun.com forgerock-sdk.local" >> C:\Windows\System32\drivers\etc\hosts   #add to new line
```
ryanbas21, line 81: windows way to edit etc/hosts file. its ugly but it gets the job done.

package.json
```diff
@@ -18,12 +18,12 @@
     "docs": "typedoc",
     "lint": "eslint --ignore-path .gitignore \"src/**/*.ts\" --fix",
     "prepublishOnly": "npm run clean:all && npm i && npm run build && npm t",
-    "start:e2e": "ws -d tests/e2e/app  -p 8443 --cors.origin null --cors.credentials --cert ./example.com+5.pem --key ./example.com+5-key.pem",
```
ryanbas21, line 21: Because we remove the certificates here, and use the ones in `lws` we can leverage the `--https` option.

package.json
```diff
@@ -60,17 +60,17 @@
...
     "jest-environment-jsdom": "^26.0.1",
     "jest-junit": "^12.0.0",
     "local-web-server": "^4.2.1",
-    "playwright": "^1.0.2",
+    "playwright": "^1.16.3",
     "prettier": "^2.1.1",
     "shx": "^0.3.2",
     "superagent": "^5.2.2",
-    "ts-jest": "^26.5.3",
+    "ts-jest": "^27.0.7",
```
ryanbas21, line 73: Upgrade here because I am leveraging the `jest/circus-runner` test runner for retries.

.github/workflows/main.yml
```diff
@@ -0,0 +1,94 @@
...
+        os: [ubuntu-18.04]
+        nodeversion: ['16']
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.nodeversion }}
+    - name: Setup custom host for sdkapp.example.com
+      run: echo "127.0.0.1 sdkapp.example.com sdks.mytestrun.com sdkapp.petrov.ca auth.example.com openig.example.com api.example.com user.example.com rest.example.com bank.example.com openam.example.com spa.example.com sdk.mytestrun.com sdks.mytestrun.com forgerock-sdk.local" | sudo tee -a /etc/hosts
```
cerebrl, line 25: I'd like to narrow this down to just the domain aliases that are needed. I think many of these could be omitted.

.github/workflows/main.yml
cerebrl: Nice job!

With all of these, let's reduce the domain aliases to just the required ones. There's a lot of legacy domains in here that I'd hate to carry it into our new process.

package.json
```diff
@@ -18,12 +18,12 @@
...
     "lint": "eslint --ignore-path .gitignore \"src/**/*.ts\" --fix",
     "prepublishOnly": "npm run clean:all && npm i && npm run build && npm t",
-    "start:e2e": "ws -d tests/e2e/app  -p 8443 --cors.origin null --cors.credentials --cert ./example.com+5.pem --key ./example.com+5-key.pem",
+    "start:e2e": "ws -d tests/e2e/app  -p 8443 --cors.origin null --cors.credentials --https",
     "start:server": "node tests/e2e/server/index.mjs",
     "start:server:live": "LIVE=true node tests/e2e/server/index.mjs",
     "start:samples": "ws -d samples -p 8443 --cors.origin null --cors.credentials --cert example.com+5.pem --key example.com+5-key.pem",
     "test": "npm run test:unit && npm run test:integration && npm run test:e2e",
-    "test:coverage": "jest --testMatch=\"<rootDir>/src/**/*.test.ts\" --config=./tests/jest.coverage.config.js --env=\"<rootDir>/tests/jest.env.config.js\"",
+    "test:coverage": "jest --testMatch=\"<rootDir>/src/**/*.test.ts\" --config=./tests/jest.basic.config.js --coverage=true",
```
cerebrl, line 26: I recently made modifications to this due to an unwanted side-effect. This would change it back to creating the side-effect. The issue was sharing the basic config was creating coverage reports for each test run, rather than just when "coverage" was run. So, our directories were getting polluted with hundreds of coverage reports.

With the update of Jest, did they fix that issue? I don't mind leaving it like this if you can ensure the same issue doesn't return.

package.json
cerebrl: I REALLY want to make sure our tests aren't flaky in the first place. If they become flaky, I'd rather fix it at the core, rather than get into the habit of rerunning them until they pass. For me, flaky tests are unacceptable.

src/http-client/helpers.test.ts
cerebrl: This is cool. Back in the day, Jest didn't want to cleanly exit when tests would finish, so I dropped `done()` within them, and it fixed it 🤷  If they now cleanly exit without it, I'm good!

tests/e2e/utilities/setup-and-go.ts
cerebrl: I'm okay with this, but I want to really stress that I want to avoid the view that E2E tests are inherently flaky. Flaky tests that can't be trusted _rapidly_ lose their value. If we see some instability in our tests, I want to evaluate the core problem and try to fix it, rather than just running the tests additional times.

cerebrl (changes requested):
The only changes I'd like to request are the removal of the unneeded domain aliases. Some of them are legacy, and I'd like to see our new implementation is clean of legacy leftovers.

For the rest, if we can ensure the coverage reports are not generated on each and every test run, our tests cleanly exit without issue, and our tests are not flaky due to these changes, I'm good with it.

package.json
ryanbas21: Good catch, I don't think I intended for this to carry over

ryanbas21: I can verify if they fixed the issue

.github/workflows/main.yml
```diff
@@ -0,0 +1,121 @@
...
+      with:
+        # npm cache files are stored in `~/.npm` on Linux/macOS
+        path: ~/.npm
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-build-${{ env.cache-name }}-
+          ${{ runner.os }}-build-
+          ${{ runner.os }}-
+    - name: Setup custom host for sdkapp.example.com
+      run: echo "127.0.0.1 sdkapp.example.com auth.example.com api.example.com openam.example.com" | sudo tee -a /etc/hosts
```
cerebrl, line 38: I don't believe any E2E tests use `openam.example.com`. This domain can be found in the codebase, but they are all unit or integration tests, which neither use a real server.

cerebrl (dismissed):
Other than the one comment, this looks good.

cerebrl (approved):

cerebrl (approved):

PR #142, opened 2021-11-12
task(setup-initial-action): Playwright Action

ryanbas21:
Rename main.yml -> develop.yml so that the checks only run against develop.

Github action filenames map to branches.


spetrov (approved):
👍🏻

PR #149, opened 2021-11-29
(docs): Sdks-1416/create-pr-template

ryanbas21:
Create a pull request template for future PR's.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository 

Documentation states that adding this file to the .github/ folder should make it so the template is used on PR's. I'm not sure if it would show up in this PR, or if we just have to hope that it shows up after its merged.

Feedback on what we want in this is welcome.

https://bugster.forgerock.org/jira/browse/SDKS-1416

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,44 @@
+# JIRA Ticket
+
+Please link jira ticket here
+
+# Description
+
+# Type of Change
+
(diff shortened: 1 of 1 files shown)


spetrov (approved):
Looks good to me!

PR #189, opened 2022-03-21
Sdks 1641/token threshold

daveadams56:
# JIRA Ticket

[SDKS-1641 [JS] Implement threshold feature in the JS SDK](https://bugster.forgerock.org/jira/browse/SDKS-1641)

# Description

Implement a configurable threshold for OAuth token expiry, within which tokens will be proactively refreshed by the SDK when tokens are used. This behaviour is implemented in the iOS and Android SDKs, but not JS. It is equivalent to the oauth_threshold property and its usage in the native SDKs - for example [oauth_threshold](https://github.com/ForgeRock/forgerock-android-sdk/blob/2983964b58268c2f91e641829c017cc2fcef87f1/forgerock-auth/src/main/res/values/strings.xml#L16) and [isExpired()](https://github.com/ForgeRock/forgerock-android-sdk/blob/2983964b58268c2f91e641829c017cc2fcef87f1/forgerock-auth/src/main/java/org/forgerock/android/auth/AccessToken.java#L67
) in the Android SDK.
# Type of Change

# How Has This Been Tested?

Unit tests have been created for the Token Manager helper function added in this change. The e2e tests have not yet been written; let's please have a discussion on the best way to approach e2e for this feature.

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [x] Unit tests are written.
- [x] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


packages/javascript-sdk/src/token-manager/helpers.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { tokensWillExpireWithinThreshold } from './helpers';
+import {
+  oauthThreshold,
+  tokenExpiryWithinThreshold,
+  tokenExpiryOutsideThreshold,
+} from './token-manager.mock.data';
+
+describe('Test Token Manager utils', () => {
+  it('Should return true if tokens expire within configured threshold', async () => {
```
ryanbas21, line 19: I don't think we need an async callback here.

packages/javascript-sdk/src/token-manager/helpers.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const expiresSoon = tokensWillExpireWithinThreshold(
+      oauthThreshold,
+      tokenExpiryWithinThreshold,
+      Date.now(),
+    );
+    expect(expiresSoon).toBe(true);
+  });
+
+  it('Should return false if tokens expire outside configured threshold', async () => {
```
ryanbas21, line 28: same as above

packages/javascript-sdk/src/token-manager/helpers.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  tokenExpiryWithinThreshold,
+  tokenExpiryOutsideThreshold,
+} from './token-manager.mock.data';
+
+describe('Test Token Manager utils', () => {
+  it('Should return true if tokens expire within configured threshold', async () => {
+    const expiresSoon = tokensWillExpireWithinThreshold(
+      oauthThreshold,
+      tokenExpiryWithinThreshold,
+      Date.now(),
```
ryanbas21, line 23: While I don't foresee a situation where using the actual Date.now() would be problematic in this scenario, I feel somewhat uneasy about using Date.now's actual value in test suite given its ever changing. 

Typically, using IO in unit tests requires mocking, and while this may be a scenario where we can get away with it, my brain has a hesitation about this. What do you think?

packages/javascript-sdk/src/token-manager/helpers.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @ignore
+ * These are private utility functions for Token Manager
+ */
+
+export function tokensWillExpireWithinThreshold(
+  oauthThreshold: number,
+  tokenExpiry: number,
+  timeNow: number,
+): boolean {
+  return tokenExpiry - oauthThreshold < timeNow;
```
ryanbas21, line 22: I believe we can reduce the API of this function since `timeNow` is always `Date.now`.  If there is a case where `timeNow` can be something other than `Date.now` perhaps we can assign it a default value of `Date.now` so we can omit this argument from the calling code unless explicitly required.

```suggestion
export function tokensWillExpireWithinThreshold(
  oauthThreshold: number,
  tokenExpiry: number,
): boolean {
  const timeNow = Date.now();
  return tokenExpiry - oauthThreshold < timeNow;
```

ryanbas21 (changes requested):
Just a few comments.

packages/javascript-sdk/src/token-manager/helpers.test.ts
daveadams56: Yes I agree it doesn't quite feel right doing it like this but was kind of on the fence. Would the alternative be to put a static timestamp in the mock data and use that for the tests?

packages/javascript-sdk/src/token-manager/helpers.ts
daveadams56: Ok I see your point. I was thinking from a testing point of view it would be better to pass in the date for predictability - that was the main reason for passing it in. If we were to go with your suggested change, how might we test this without having Date.now() in our tests?

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -170,10 +170,16 @@ abstract class OAuth2Client {
       throw new Error('Access token not found in response');
     }
 
+    if (!responseObject.expires_in) {
+      throw new Error('Expires in not found in response');
+    }
+    const tokenExpiry = Date.now() + responseObject.expires_in / 1000;
```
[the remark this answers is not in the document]
daveadams56: Yes, this does seem more correct and readable

packages/javascript-sdk/src/token-manager/helpers.test.ts
daveadams56: Good point!

daveadams56:
I know we need to discuss how to implement e2e tests for this, but aside from this I've noticed that the e2e run on push seems to fail: `Error: Cannot find module './teardown'`

I've been pushing by skipping this step. Any idea why this would be failing?

packages/javascript-sdk/src/token-manager/helpers.ts
ryanbas21: I think you have to mock the the Date object. Something like

```
describe('Test', () => {
  // to avoid race conditions between expected and actual date values
  beforeAll(() => {
    jest.useFakeTimers('modern'); // tell Jest to use a different timer implementation.
    jest.setSystemTime(new Date('20 Aug 2020 00:12:00 GMT').getTime())
  });

  afterAll(() => {
    // Back to reality...
    jest.useRealTimers();
  });
  ```
  https://github.com/facebook/jest/issues/2234#issuecomment-731451275

**Comment by ryanbas21** [comment]
> I know we need to discuss how to implement e2e tests for this, but aside from this I've noticed that the e2e run on push seems to fail: `Error: Cannot find module './teardown'`
> 
> I've been pushing by skipping this step. Any idea why this would be failing?

rebase with develop (also messaged in slack).

File: packages/javascript-sdk/src/token-manager/helpers.ts (hunk 4 above)
Comment (daveadams56) [reply packages/javascript-sdk/src/token-manager/helpers.ts:22]: Makes sense, I'll have a go at mocking the date. Thanks

**Review by cerebrl — CHANGES_REQUESTED** [review]
A few pattern change suggestions on one of the modules, but I like the overall direction. Good job!

File: packages/javascript-sdk/src/oauth2-client/index.ts [hunk 6]
```diff
@@ -170,10 +170,16 @@ abstract class OAuth2Client {
       throw new Error('Access token not found in response');
     }
 
+    if (!responseObject.expires_in) {
+      throw new Error('Expires in not found in response');
```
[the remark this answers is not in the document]
daveadams56: I guess it may be over the top, since having no expires_in just means that we don't know when they expire, so we can't pro-actively refresh them (as before this change). I'll look into alternatives 👍

packages/javascript-sdk/src/config/interfaces.ts
```diff
@@ -33,6 +33,7 @@ interface ConfigOptions {
   tokenStore?: TokenStoreObject | 'indexedDB' | 'sessionStorage' | 'localStorage';
   tree?: string;
   type?: string;
+  oauthThreshold?: number;
```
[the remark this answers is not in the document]
daveadams56: Yes I agree with you here, but I went with consistency with the iOS and Android SDKs - they both call it **forgerock_oauth_threshold** in the config. If you or @ryanbas21 feel strongly that we should go with something more descriptive then I'm happy to go with that, otherwise we can keep it somewhat consistent

packages/javascript-sdk/src/config/index.ts
```diff
@@ -37,6 +37,8 @@ abstract class Config {
    * @param options The options to use as defaults
    */
   public static set(options: ConfigOptions): void {
+    this.setDefaults(options);
```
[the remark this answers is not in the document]
daveadams56: This is really helpful - thank you! I love getting examples for stuff like this, you've made it really understandable.

I've stepped through the examples you've given above and I agree with what you're saying. The only thing I'm not sure of now is where the function setDefaults should live. Having it in the same file would work I think but triggers a linting warning when it's used: `'setDefaults' was used before it was defined.`. Maybe we should move it out to a utility file after all?

packages/javascript-sdk/src/oauth2-client/index.ts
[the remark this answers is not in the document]
daveadams56: Just noticed that **prettier** is complaining about the parentheses - I think that's probably why I omitted them in the first place. @ryanbas21 what do you think - shall we live with the prettier warning, or add an exception - or change it back?
Screenshot 2022-03-24 at 21 23 04

packages/javascript-sdk/src/config/index.ts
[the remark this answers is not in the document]
cerebrl: Cool, glad it was helpful :)

You can fix that lint warning by moving the function definition _above_ the class within which it is used. Using a function before it's defined will usually work because of function hoisting, but it's generally considered a bad practice. So, linters will warn you when you do it.

packages/javascript-sdk/src/oauth2-client/index.ts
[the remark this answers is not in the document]
cerebrl: We _should_ be able to turn off that warning. I'm not sure I even agree with it 🤔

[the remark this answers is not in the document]
daveadams56: For now I've just gone with `// eslint-disable-next-line` as I'm not sure how else we might turn off the warning

packages/javascript-sdk/src/config/index.ts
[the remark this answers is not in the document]
daveadams56: Makes sense, I've reworked this thanks to your advice

packages/javascript-sdk/src/token-manager/helpers.ts
daveadams56: This worked easily enough - thanks 👍

packages/javascript-sdk/src/token-manager/helpers.test.ts
daveadams56: I've removed the date from here, feels more right. Thanks

ryanbas21 (approved):
Looks good from my perspective, just would like the arithmetic parenthesized so its clear to read due to operator precedence.

👍 nice!

edit: It appears prettier removes the parens i'm asking for so we shall go with what prettier says!

packages/javascript-sdk/src/config/index.ts
```diff
@@ -43,7 +55,9 @@ abstract class Config {
     if (options.serverConfig) {
       this.validateServerConfig(options.serverConfig);
     }
-    this.options = { ...options };
+    this.options = {
+      ...setDefaults(options),
+    };
```
cerebrl, line 60: Clean!

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -170,10 +170,16 @@ abstract class OAuth2Client {
       throw new Error('Access token not found in response');
     }
 
+    let tokenExpiry: number | undefined = undefined;
+    if (responseObject.expires_in) {
+      tokenExpiry = (Date.now() + responseObject.expires_in) / 1000;
```
cerebrl, line 175: Ah, good catch!

cerebrl (commented):
This is starting to shape up well. Let's wrap it up with some E2E tests and then squash all the commits to get this merged.

e2e/autoscript-apps/src/authn-oauth/autoscript.ts
```diff
@@ -111,6 +116,24 @@ function autoscript() {
           return newTokens;
         },
       ),
+
+      rxDelay(delay),
+      mergeMap(
+        (tokens) => {
+          console.log('Proactively refresh tokens if expiring soon');
+          rxDelay(delay);
+          return forgerock.TokenManager.getTokens();
```
ryanbas21, line 125: I'm not an rxjs expert - and I think this is technically correct - but my functional brain tells me that this type should be an observable.

While I know promises and observables interop, I think this should be an `of(forgerock.TokenManager.getTokens()`. 

It may be a moot point, i'm not 100% sure so maybe @cerebrl can correct me here. I just know that typically monadic interfaces return the same context (observable) in this case.

e2e/autoscript-apps/src/authn-oauth/autoscript.ts
```diff
@@ -100,6 +103,8 @@ function autoscript() {
       mergeMap(
         (tokens) => {
           console.log('Force renew OAuth tokens');
+          // Pause for 2 seconds to allow expired tokens to expire
+          rxDelay(tokenExpiredDelay);
```
ryanbas21, line 107: same as below I think? @cerebrl maybe you can correct me here

ryanbas21 (changes requested):
Just a few thing i'm unsure of.

e2e/autoscript-apps/src/authn-oauth/autoscript.ts
cerebrl, line 107: I'm assuming this works, but I would lose the bet if I was asked if it did. Being that the `rxDelay` is not "attached" to any event "stream", I would think the JS engine would evaluate it, but then just move on and call the `getTokens` method without waiting for the delay's completion event.

What I would expect to see is a use of the `rxDelay` on line 102 that's integrated into the main event stream. Was there a reason you didn't utilize the delay just above this `mergeMap`?

e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
```diff
@@ -31,4 +31,51 @@ test.describe('Test OAuth login flow', () => {
...
+  test(`should not proactively refresh tokens if they expire outside specified threshold`, async ({
+    page,
+    browserName,
+  }) => {
+    const { messageArray } = await setupAndGo(page, browserName, 'authn-oauth/', {
+      realmPath: 'tokens-expiring-soon',
+      oauthThreshold: '10000',
+    });
+    expect(messageArray.includes('OAuth tokens not expiring soon; not refreshed')).toBe(true);
+  });
```
cerebrl, line 73: There are 5 newly introduced tests here, which means 5 independent browser launches. Can we reuse the same browser lunch for one or more of these? We should be able to, at least, move this one into the original OAuth test, yeah? Having independent browser launches if not bad, per se, but it does come with a performance liability at scale.

e2e/mock-api/src/app/routes.auth.js
```diff
@@ -289,6 +291,12 @@ export default function (app) {
       } else {
         res.status(406).send('Middleware header is missing.');
       }
+    } else if (req.path.includes('tokens-expiring-soon')) {
+      const tokensExpiringSoon = { ...oauthTokensExpiringSoon, access_token };
+      res.json(tokensExpiringSoon);
+    } else if (req.path.includes('tokens-expired')) {
+      const tokensExpired = { ...oauthTokensExpired, access_token };
+      res.json(tokensExpired);
```
cerebrl, line 299: Just to make sure, object spreads evaluate from left to right. So, if the `access_token` object and the `oauthTokensExpir*` object being spread contain the same property, `access_token` will currently have priority. Is that what you're intending here? If `access_token` is the default object, and `oauthTokensExpi*` is the "modifying" object, then I'd put `access_token` first (left).

cerebrl (commented):
A few comments to ensure e2e testing is optimized.

e2e/autoscript-apps/src/authn-oauth/autoscript.ts
cerebrl: This is following a pattern heavily used throughout our autoscripts, so I wouldn't worry about it. But yes, Ryan, you are likely correct from a technical sense. From a practical one, this works without issue.

daveadams56: Ok, I understand what you mean now. I went with the pattern throughout the rest of the tests, I thought it's probably best to be consistent if this works.

e2e/autoscript-apps/src/authn-oauth/autoscript.ts
```diff
@@ -111,6 +116,24 @@ function autoscript() {
           return newTokens;
         },
       ),
+
+      rxDelay(delay),
+      mergeMap(
+        (tokens) => {
+          console.log('Proactively refresh tokens if expiring soon');
+          rxDelay(delay);
```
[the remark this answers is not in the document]
daveadams56: Yes, you're entirely right. I required a 2 second delay to let the "expiring in 1 second" tokens expire in order to exercise the case where the tokens had expired and **getTokens()** is called. In doing so, I got a little confused about the effect **rxDelay** would have in the test and was thinking along the lines of a **setTimeout**. I think the "tokens expired" test case passed despite this. I've now changed the test so that the **rxDelay** is just above the **mergeMap** and I can now see that it works correctly.

e2e/autoscript-apps/src/authn-oauth/autoscript.ts
daveadams56: This delay was added by mistake, and will be removed

daveadams56: I think I was a little confused about how the rxDelay worked. This has been corrected.

e2e/mock-api/src/app/responses.js
```diff
@@ -18,6 +18,22 @@ export const oauthTokens = {
...
+};
+
+export const oauthTokensExpired = {
+  access_token: 'baz',
+  scope: 'openid profile me.read',
+  id_token: 'mox',
+  token_type: 'Bearer',
+  expires_in: 1,
+};
+
```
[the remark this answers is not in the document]
daveadams56: Yes, hadn't thought of this but good idea.

e2e/mock-api/src/app/routes.auth.js
daveadams56: Yes that is the intention, for consistency with the rest of the function that returns the regular **oauthTokens** response with access_token replacing the value within the oauthResponse. As per the above conversation, it looks like v4() replaces the value of the access token. Does this sound correct to you? Or am I replacing it when I don't need to?

e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
daveadams56: Ok, I see your point about multiple browser launches. I think we can combine a few of these by having multiple **setupAndgo** calls within each **test** block. I'll have a go at rearranging these accordingly.

ryanbas21 (approved):
I approve pending the rest of your changes from @cerebrl review.

Great work Dave!

e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
daveadams56: I've grouped the OAuth tests into three browser launches - one for each type of test outcome as it felt like it made sense. I could however reduce further and just have one launch for the whole test script. What do you think?

e2e/mock-api/src/app/responses.js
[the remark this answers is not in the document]
cerebrl: Oh, you are correct. Sorry for the confusion. This also means my other comment can be ignored related to object spread, I think.

e2e/mock-api/src/app/routes.auth.js
cerebrl: Yes, you are correct. I don't know why, but I confused `access_token` with a tokens object, rather than what it is, a string value. Sorry!

e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
cerebrl: It's a judgement call, and I'll leave it up to you to figure out what's the right number. I don't want to us to be overly concerned about separate browser launches, but I just want us to be mindful.

daveadams56: Understood. Thanks

cerebrl (approved):

PR #267, opened 2022-07-27
ci: nx-affected-fix

ryanbas21:
the affected commands require the `version` target to be a cacheable
operation in the pipeline

https://nx.app/runs/bXFJ7UcofMQ to see it working

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -5,7 +5,7 @@
     "default": {
       "runner": "@nrwl/nx-cloud",
       "options": {
-        "cacheableOperations": ["build", "test", "lint", "package", "prepare", "e2e"],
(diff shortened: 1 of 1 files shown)


daveadams56 (commented):
This makes sense - I'm guessing this is something that did not necessarily show up until the previous PR was merged (and something failed), have I got this right?

daveadams56 (commented):
Nx run looks good, as do the deployed sample apps. Happy to approve once I have confirmation of my assumption on what this is (see other comment)

daveadams56 (approved):
Discussed directly - looks good to me. Approved

PR #299, opened 2022-12-07
SDKS-2195 Update vars in samples to match AM setup instructions.

forgerock-chris:
# JIRA Ticket

[SDKS-2195](https://bugster.forgerock.org/jira/browse/SDKS-2195)

# Description

Updated the example values for configuring the JavaScript samples to match each other, and the values we use in other tutorials/setting up docs.

# Type of Change

Minor usability change

# How Has This Been Tested?

PR review

# Definition of Done

Check all that apply

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


samples/angular-todo/src/app/app.component.ts
```diff
@@ -56,21 +56,21 @@ export class AppComponent implements OnInit {
...
+     *   include the deployment path at the end, such as `/am/` or `/openam/`
+     * - realmPath: this is the realm to use within ForgeRock. such as `alpha` or `root`
+     * - tree: The authentication journey/tree to use, such as `sdkAuthenticationTree`
      *************************************************************************** */
 
     Config.set({
       clientId: environment.WEB_OAUTH_CLIENT,
       redirectUri: `${window.location.origin}/callback`,
-      scope: 'openid profile email',
+      scope: 'openid profile email address phone',
```
daveadams56, line 73: Why are we adding the `address` and `phone` scopes here?

samples/embedded-login/src/main.js
```diff
@@ -13,14 +13,14 @@ import * as forgerock from '@forgerock/javascript-sdk';
 const FATAL = 'Fatal';
 forgerock.Config.set({
   clientId: process.env.WEB_OAUTH_CLIENT, // e.g. 'ForgeRockSDKClient'
-  redirectUri: `${window.location.origin}/callback`, // e.g. 'https://sdkapp.example.com:8443/_callback'
-  scope: process.env.SCOPE, // e.g. 'openid profile me.read'
+  redirectUri: `${window.location.origin}/callback`, // e.g. 'https://sdkapp.example.com:8443/callback'
+  scope: process.env.SCOPE, // e.g. 'openid profile email address phone me.read'
   serverConfig: {
-    baseUrl: process.env.AM_URL, // e.g. 'https://openam.example.com:9443/openam/'
+    baseUrl: process.env.AM_URL, // e.g. 'https://myorg.forgeblocks.com/am' or 'https://openam.example.com:8443/openam/'
```
daveadams56, line 19: The first example URL is /am and the second is /openam/ - I don't think it matters to the samples whether there is one or not, but better to be consistent?

daveadams56 (commented):
Looks good but I just want to understand why we are adding scopes to the samples before I approve.

samples/angular-todo/src/app/app.component.ts
forgerock-chris: Purely to match the setup we ask the readers to do when creating the OAuth 2.0 client. I suspect we don't even use them all, but wanted them to (perhaps subconsciously) marry the two steps together.

daveadams56: Ok that's fine - just wanted to be sure that the reader will probably have configured the OAuth 2.0 client accordingly. Thanks.

daveadams56 (approved):

samples/angular-todo/src/app/app.component.ts
cerebrl: What setup are we aligning to here? The [blog article](https://backstage.forgerock.com/docs/sdks/latest/blog/build-protected-angular-web-app.html#public_aka_spa_type_oauth_client_settings) that this sample app corresponds with instructs the reader to only add the first three. If you are referring to the [tutorial from the core docs](https://backstage.forgerock.com/docs/sdks/latest/javascript/tutorial.html), that's the other style of "sample" apps. If we want to align all of the sample apps, we'd have to update quite a few more places in other branches as well.

samples/embedded-login/.env.example
```diff
@@ -1,7 +1,6 @@
 AM_URL=$AM_URL
-CLIENT_ID=$CLIENT_ID
+WEB_OAUTH_CLIENT=$WEB_OAUTH_CLIENT
```
cerebrl, line 2: I'd like to keep these alphabetized, if possible. It just makes it easier to find variables when the list grows.

samples/embedded-login/src/main.js
```diff
@@ -13,14 +13,14 @@ import * as forgerock from '@forgerock/javascript-sdk';
 const FATAL = 'Fatal';
 forgerock.Config.set({
   clientId: process.env.WEB_OAUTH_CLIENT, // e.g. 'ForgeRockSDKClient'
-  redirectUri: `${window.location.origin}/callback`, // e.g. 'https://sdkapp.example.com:8443/_callback'
-  scope: process.env.SCOPE, // e.g. 'openid profile me.read'
+  redirectUri: `${window.location.origin}/callback`, // e.g. 'https://sdkapp.example.com:8443/callback'
+  scope: process.env.SCOPE, // e.g. 'openid profile email address phone me.read'
```
cerebrl, line 17: It looks like we don't recommend the `me.read` scope any more in our core tutorial docs: https://backstage.forgerock.com/docs/sdks/latest/serverconfiguration/cloud/create-oauth2-client.html.

cerebrl (commented):
Just a few comments, but looks good overall. One of the failures is the commit message syntax you are using. It's not meeting the requirements seen here: https://www.conventionalcommits.org/en/v1.0.0-beta.4/#summary. I'll get Ryan to help with a good example of what will pass.

samples/central-login/.env.example
```diff
@@ -1,6 +1,5 @@
 AM_URL=$AM_URL
 REALM_PATH=$REALM_PATH
-REDIRECT_URI=$REDIRECT_URI
```
ryanbas21, line 3: Do we not use this in blog posts?

forgerock-chris: It doesn't seem to be used anywhere in the sample - is it required?

samples/embedded-login/src/main.js
forgerock-chris: Yeah I left that alone in case it was needed by the samples, but if not let's pull that too.

samples/embedded-login/.env.example
forgerock-chris: Good call, will fix.

samples/angular-todo/src/app/app.component.ts
forgerock-chris: Yeah as you guessed I was aligning to the "Configure the Service" documentation, and hadn't noticed there was a blog post that corresponds to the official samples.
My goal is to unify all server-side setup in a single place, but have it cover all the samples and tutorials on the site. In the meantime, I'll remove the additional scopes from the code. If they use the blog to setup they'll have them, if they use the docs they'll have a few spare scopes in the client, that's all.

samples/central-login/.env.example
daveadams56: It's not used in the central-login sample, though one of the files ([webpack.config.js](https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/samples/central-login/webpack.config.js#L7)), references it. Omitting this from the .env doesn't seem to cause any issues with the central-login app in any case, so it should probably be removed from ([webpack.config.js](https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/samples/central-login/webpack.config.js#L7)).

forgerock-chris: I've removed it from there also, thanks David.

forgerock-chris:
Going to close this one as I branched from master rather than develop, and had a commit I didn't want involved.
Have opened a new PR after cherry-picking this across to develop which ought to be cleaner.
Thanks for looking!

PR #300, opened 2022-12-15
docs: updated vars in samples to match AM setup instructions

forgerock-chris:
fixes SDKS-2195

Recreated original PR as I accidentally had an incorrect commit included (I think I branched from master, rather than develop, apologies!)


forgerock-chris:
Should not have been from a fork. Closing the PR.

PR #308, opened 2023-01-12
feat: process policy advice as json

daveadams56:
# JIRA Ticket

Please link jira ticket here

[SDKS-2238: [JS SDK] Process IG policy advice as JSON](https://bugster.forgerock.org/jira/browse/SDKS-2238)

# Description

Transaction Authorization advice information is usually returned by IG as query parameters in a http redirect but [OPENIG-6642](https://bugster.forgerock.org/jira/browse/OPENIG-6642) adds the support for JSON advice - this branch modifies the HttpClient within the JS SDK to be compatible with JSON based advice.

# Type of Change

- [x] This change requires a documentation update

# How Has This Been Tested?

Additional e2e test suite added for JSON advice - see [authz-txn-basic-json](https://github.com/daveadams56/forgerock-javascript-sdk/tree/feat_process-json-advice/e2e/autoscript-apps/src/authz-txn-basic-json) and [suites/authz-txn-basic-json.test.ts](https://github.com/daveadams56/forgerock-javascript-sdk/blob/feat_process-json-advice/e2e/autoscript-suites/src/suites/authz-txn-basic-json.test.ts).

Some testing against a live environment has been performed using the e2e apps. See the JIRA ticket for environment details if you want to test against Stoyan's test environment.

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [x] Unit tests are written.
- [x] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


e2e/autoscript-suites/src/suites/authz-txn-basic-redirect.test.ts
```diff
@@ -1,7 +1,7 @@
 /*
  * @forgerock/javascript-sdk
  *
- * authz-txn-basic.test.ts
+ * authz-txn-basic-redirect.test.ts
```
daveadams56, line 4: I decided to split out the authz-txn-basic test into two - one handling advice as redirect, the other as JSON

e2e/mock-api/src/app/responses.js
```diff
@@ -63,6 +63,20 @@ export const createStepUpUrl = (url) => {
   return redirectUrl.toString();
 };
 
+export const createStepUpHeader = (url) => {
+  // Grab the client's desired AM URL
+  const referer = new URL(url);
+  const amUrl = referer.searchParams.get('amUrl') || AM_URL;
+
+  // Base 64 of {"TransactionConditionAdvice":["39dfdd15-59a3-473c-a7fc-ecda3bbc3bc8"]}
```
daveadams56, line 71: Kept txid the same as the one we use elsewhere in the tests

e2e/mock-api/src/app/responses.js
```diff
@@ -63,6 +63,20 @@ export const createStepUpUrl = (url) => {
...
+  const advices =
+    'eyJUcmFuc2FjdGlvbkNvbmRpdGlvbkFkdmljZSI6WyIzOWRmZGQxNS01OWEzLTQ3M2MtYTdmYy1lY2RhM2JiYzNiYzgiXX0=';
+  const format = '1';
```
daveadams56, line 74: So far in my live testing I have only seen "format=1" - any idea what this means?

e2e/mock-api/src/app/routes.resource.js
```diff
@@ -115,7 +120,13 @@ export default function (app) {
         baz.canWithdraw = false;
         res.json({ message: 'Successfully retrieved resource!' });
       } else {
-        res.redirect(307, createStepUpUrl(req.headers.referer));
+        if (req.headers['x-forgerock-sdk']) {
+          res.setHeader('WWW-Authenticate', createStepUpHeader(req.headers.referer));
+          res.setHeader('Access-Control-Expose-Headers', 'www-authenticate');
```
daveadams56, line 125: I had to add this in order for the test app to be able to view the WWW-Authenticate header on 401 response - otherwise I seemed to have no access to headers, which was weird. IG provides this header by the way, so I thought it was fine to include here.

packages/javascript-sdk/src/http-client/index.ts
```diff
@@ -28,11 +28,13 @@ import {
   addAuthzInfoToURL,
   buildAuthzOptions,
   examineForIGAuthz,
+  examineForIGAuthzHeader,
   examineForRESTAuthz,
   hasAuthzAdvice,
   isAuthzStep,
   newTokenRequired,
```
daveadams56, line 35: This import of newTokenRequired (and the subsequent usage), was removed in the PoC code. @cerebrl can you remember why this was? I kept it in as I couldn't see how this affects the flow.

packages/javascript-sdk/src/http-client/index.ts
```diff
@@ -167,20 +172,15 @@ abstract class HttpClient extends Dispatcher {
...
+  private static async stepIterator(res: Response, handleStep: HandleStep): Promise<void> {
     const jsonRes = await res.json();
     const initialStep = new FRStep(jsonRes);
 
     // eslint-disable-next-line no-async-promise-executor
     return new Promise(async (resolve, reject) => {
       async function handleNext(step: FRStep): Promise<void> {
         const input = await handleStep(step);
-        const output = await FRAuth.next(input, { type, tree });
+        const output = await FRAuth.next(input, { tree: '', type: '' });
```
daveadams56, line 183: I eventually figured out that in order for the transaction approval tree-outcome to take effect, we could only pass the `authIndexType=composite_advice` on the initial authentication request. Having any query parameters for `authIndexType` or `authIndexValue` after the first one, caused a 401 described as "session expired". It was weird but I found a conversation @cerebrl had with others in the team, where this was discovered. I'm fine having this in here, but we may want to discuss why it is necessary for this flow now when previously that may not have been the case - @cerebrl.

e2e/autoscript-apps/src/authz-txn-basic-json/autoscript.ts
```diff
@@ -0,0 +1,140 @@
...
+      rxDelay(delay),
+      mergeMap(
+        (step) => {
+          console.log('Retrieve the protected resource');
+          return forgerock.HttpClient.request({
+            url: `${igUrl ? igUrl : resourceOrigin + '/ig/authz-by-txn'}`,
+            init: {
+              method: 'GET',
+              credentials: 'include',
+              headers: { 'X-ForgeRock-SDK': true },
```
daveadams56, line 93: This is the new header that we need to pass to prompt IG to return the advice as JSON, in case you missed it.

e2e/autoscript-apps/src/authz-txn-basic-redirect/autoscript.ts
```diff
@@ -39,10 +39,7 @@ function autoscript() {
             }
             break;
           case 'AUTHENTICATE':
-            if (
-              action.payload.type === 'composite_advice' &&
-              typeof action.payload.tree === 'string'
-            ) {
+            if (action.payload.tree === '') {
```
daveadams56, line 42: This is necessary for the tests to pass, now that we are only passing authIndexType=composite_advice (or indeed any query parameters at all), on the initial authentication request when we act upon the advice returned by the server. An action type of AUTHENTICATE and no tree specified is about as specific as I can be here, I think, when determining which stage of the test we are at for the suite.

packages/javascript-sdk/src/http-client/index.ts
cerebrl: I commented it out, but I don't remember why. I was likely just for ease of testing; definitely bring it back :)

packages/javascript-sdk/src/http-client/index.ts
cerebrl: Hey David, sorry for not properly documenting this. I remember writing it down somewhere, but I can't, for the life of me, remember where. You are correct. Something in AM has changed that made it much more sensitive to these parameters. If I remember correctly, the fix is removing the `tree` property from the root SDK config. That way, only when a `tree` value is explicitly passed in does it add those parameters.

e2e/autoscript-apps/src/authz-txn-basic-json/autoscript.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  const url = new URL(window.location.href);
+  const amUrl = url.searchParams.get('amUrl') || 'https://auth.example.com:9443/am';
+  const realmPath = url.searchParams.get('realmPath') || 'root';
+  const igUrl = url.searchParams.get('igUrl'); // only use when testing against IG on different host
+  const resourceOrigin =
+    url.searchParams.get('resourceOrigin') || 'https://api.example.com:9443/resource';
+  const un = url.searchParams.get('un') || 'sdkuser';
+  const pw = url.searchParams.get('pw') || 'password';
+  const tree = url.searchParams.get('tree') || 'UsernamePassword';
```
cerebrl, line 26: Removing this default _should_ prevent the `authIndex*` issue brought up in the other comment.

e2e/autoscript-apps/src/authz-txn-basic-json/autoscript.ts
cerebrl: As discussed in the Friday call, this will need to be moved to the SDK itself. But, we need to verify that this is indeed the "official" way to communicate to IG that a redirect is not wanted. I'm going to add you to a Slack channel for discussing things like this.

packages/javascript-sdk/src/http-client/index.ts
daveadams56: Ok, so are you saying that we do not need this code change - and instead require our tests (and therefore the users of the SDK), to remove the tree property from the root config in order for the tree to be omitted and this flow to work? I'm fine with removing it from our tests but would be more concerned about pushing developers into configuring the SDK a certain way for this to work. I guess if we at least document this then it's ok, but perhaps we should think about raising a ticket on AM or adapting the SDK similar to the above change. What do you think?

e2e/mock-api/src/app/responses.js
cerebrl: I think `format=1` is the only format relevant to us. If I remember correctly, the other formats are for contexts the SDK doesn't handle.

cerebrl: Though, explicitly asking this in the channel to which I just added you might be a good idea.

e2e/mock-api/src/app/routes.resource.js
cerebrl: This should likely go into the CORS configuration here: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/e2e/mock-api/src/index.js#L31. Anything related to `Access-Control` is a CORS concern.

cerebrl: Docs for the CORS middleware can be found here: https://expressjs.com/en/resources/middleware/cors.html

e2e/autoscript-apps/src/authz-txn-basic-json/autoscript.ts
daveadams56: Thanks - I can see the slack channel now. I had a chat with the IG team yesterday - let me update you on slack about the status of the IG work.

cerebrl (commented):
Left a few comments and responded to a few of yours. Overall, it's looking good. I didn't see any "authorize by tree" testing. Is that just not yet implemented?

e2e/mock-api/src/app/routes.resource.js
daveadams56: That makes a lot more sense. Thanks!

packages/javascript-sdk/src/http-client/index.ts
daveadams56: Ok no worries, I have kept it in there guessing this might be the case.

e2e/autoscript-apps/src/authz-txn-basic-json/autoscript.ts
daveadams56: Hmm ok I'll give that a try.

daveadams56:
> Left a few comments and responded to a few of yours. Overall, it's looking good. I didn't see any "authorize by tree" testing. Is that just not yet implemented?

Great, thanks for your feedback. Good point on the "by tree" advice - I had intended to do this at some point once we've reviewed the approach for transactional authz but forgot to say. I'll get the tests implemented along with your suggestions. Thanks.

e2e/mock-api/src/app/responses.js
daveadams56: The IG team have replied to say where this came from - looks like it was something you suggested in your gist before the PoC: https://gist.github.com/cerebrl/30e0da4b21a8b7b0eb6e2eed94279a7b

packages/javascript-sdk/src/http-client/index.ts
cerebrl: I'm going to write up a story to improve the core module of the SDK. That way we can track this. As far as this particular use case, you may have to preserve this fix, but I'd like to address the underlying issue, rather than patch this symptom.

packages/javascript-sdk/src/http-client/helpers.ts
```diff
@@ -182,6 +202,19 @@ export function normalizeIGJSON(res: Response): AuthorizationJSON {
   };
 }
 
+export function normalizeIGJSONResponseToAdviceJSON(res: Response): AuthorizationJSON {
```
ryanbas21, line 205: I think we should probably just accept the header needed in this function. We don't need to leak the entire response into this function since this only operates on headers.

I think we can therefore create a type that checks specifically for the WWW-Authenticate header.

```ts
type AuthenticateHeader = "WWW-Authenticate"
```

ryanbas21: this may not be as possible as i thought since repsonse headers are just strings and can't be narrowed from what i can tell

ryanbas21: however we can still narrow the scope of the parameter.

e2e/autoscript-apps/src/authz-tree-basic-redirect/autoscript.ts
```diff
@@ -107,7 +104,7 @@ function autoscript() {
           if (response.ok) {
             console.log('Request to IG resource successfully responded');
           } else {
-            throw new Error('IG Transactional Authorization was not successful');
+            throw new Error('IG Tree Based Authorization was not successful');
```
daveadams56, line 107: If you're scrolling through the changes from the top, this is the first change that you can see where I'm fixing something on the authz-tree-basic tests that was referencing **Transactional Authorization** instead of **Tree Based Authorization**. I'll refer to test fixes for transactional / tree-based authorization in changes further down.

e2e/autoscript-suites/src/suites/authz-tree-basic-redirect.test.ts
```diff
@@ -11,9 +11,9 @@
 import { test, expect } from '@playwright/test';
 import { setupAndGo } from '../utilities/setup-and-go';
 
-test.describe('Test Transaction Authorization flow', () => {
-  test(`Trigger Txn Auth appropriately`, async ({ page, browserName }) => {
-    const { messageArray } = await setupAndGo(page, browserName, 'authz-txn-basic/');
+test.describe('Test Tree Based Authorization flow using Redirect response', () => {
+  test(`Trigger Tree Based Auth appropriately`, async ({ page, browserName }) => {
+    const { messageArray } = await setupAndGo(page, browserName, 'authz-tree-basic-redirect/');
```
daveadams56, line 16: This is an example of the test fixes for transactional / tree-based authorization that I feel are necessary; here we had the Tree Based test referencing the setup for the Transactional test app. This means that certain responses would not have been tested.

e2e/mock-api/src/app/responses.js
```diff
@@ -63,6 +63,52 @@ export const createStepUpUrl = (url) => {
   return redirectUrl.toString();
 };
 
+export const createTreeStepUpUrl = (url) => {
```
daveadams56, line 66: As part of the test fixes for transactional / tree-based authorization I decided to split a few things out based on whether they are creating headers or URLs for tree-based or transactional authorization use cases.

packages/javascript-sdk/src/http-client/index.ts
daveadams56: Ok I think that makes sense. Thanks for creating the story - I'll keep this code as-is for now.

packages/javascript-sdk/src/http-client/helpers.ts
daveadams56: Narrowing the scope of the parameter makes sense - I'll push this in a bit. Thanks for the suggestion.

daveadams56:
I'm closing this PR since this story is now on hold pending implementation from the IG team. The branch will be retained and a further PR opened when this is picked up again.

PR #329, opened 2023-02-09
fix(javascript-sdk): use-prompt-none

ryanbas21:
use the prompt=none query parameter.

# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2060

# Description
prompt none provides a better experience when errors occur in an iframe during /authorize

# Type of Change

# How Has This Been Tested?

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread

### DIFF
--- a/packages/javascript-sdk/src/oauth2-client/index.ts
+++ b/packages/javascript-sdk/src/oauth2-client/index.ts
@@ -38,6 +38,9 @@ const allowedErrors = {
   NetworkError: 'NetworkError when attempting to fetch resource.',
   // Webkit browser error
   CORSError: 'Cross-origin redirection',
+
+  //Prompt error
+  PromptError: 'User is not authenticated or session cookie not sent to the server',
 };
 
 /**
@@ -50,6 +53,7 @@ abstract class OAuth2Client {
     const requestParams: StringDict<string | undefined> = {
       ...options.query,
       client_id: clientId,
+      prompt: options.login ? options.login : 'none',
       redirect_uri: redirectUri,
       response_type: options.responseType,
       scope,
@@ -158,6 +162,12 @@ abstract class OAuth2Client {
     const responseBody = await this.getBody<unknown>(response);
 
     if (response.status !== 200) {
+      const error_description_query_param =
+        new URL(response.url).searchParams.get('error_description') !== undefined;
+
+      if (error_description_query_param) {
+        throw new Error(allowedErrors.PromptError);
+      }
       const message =
         typeof responseBody === 'string'
           ? `Expected 200, received ${response.status}`
--- a/packages/javascript-sdk/src/oauth2-client/index.ts
+++ b/packages/javascript-sdk/src/oauth2-client/index.ts
@@ -39,8 +39,8 @@ const allowedErrors = {
   // Webkit browser error
   CORSError: 'Cross-origin redirection',
 
-  //Prompt error
-  PromptError: 'User is not authenticated or session cookie not sent to the server',
+  // prompt=none errors
+  InteractionNotAllowed: 'The request requires some interaction that is not allowed.',
 };
 
 /**
@@ -49,15 +49,14 @@ const allowedErrors = {
 abstract class OAuth2Client {
   public static async createAuthorizeUrl(options: GetAuthorizationUrlOptions): Promise<string> {
     const { clientId, middleware, redirectUri, scope } = Config.get(options);
-
     const requestParams: StringDict<string | undefined> = {
       ...options.query,
       client_id: clientId,
-      prompt: options.login ? options.login : 'none',
       redirect_uri: redirectUri,
       response_type: options.responseType,
       scope,
       state: options.state,
+      ...(options.prompt ? { prompt: options.prompt } : {}),
     };
 
     if (options.verifier) {
@@ -86,7 +85,8 @@ abstract class OAuth2Client {
    * New Name: getAuthCodeByIframe
    */
   public static async getAuthCodeByIframe(options: GetAuthorizationUrlOptions): Promise<string> {
-    const url = await this.createAuthorizeUrl(options);
+    const url = await this.createAuthorizeUrl({ ...options, prompt: 'none' });
+
     const { serverConfig } = Config.get(options);
 
     return new Promise((resolve, reject) => {
@@ -162,12 +162,6 @@ abstract class OAuth2Client {
     const responseBody = await this.getBody<unknown>(response);
 
     if (response.status !== 200) {
-      const error_description_query_param =
-        new URL(response.url).searchParams.get('error_description') !== undefined;
-
-      if (error_description_query_param) {
-        throw new Error(allowedErrors.PromptError);
-      }
       const message =
         typeof responseBody === 'string'
           ? `Expected 200, received ${response.status}`
--- a/packages/javascript-sdk/src/oauth2-client/interfaces.ts
+++ b/packages/javascript-sdk/src/oauth2-client/interfaces.ts
@@ -11,7 +11,6 @@
 import { StringDict } from '../shared/interfaces';
 import { ConfigOptions } from '../config';
 import { ResponseType } from './enums';
-import { GetTokensOptions } from '../token-manager';
 
 /**
  * Tokens returned after successful authentication.
@@ -36,11 +35,12 @@ interface AccessTokenResponse {
 /**
  * Options used when requesting the authorization URL.
  */
-interface GetAuthorizationUrlOptions extends GetTokensOptions {
+interface GetAuthorizationUrlOptions extends ConfigOptions {
   responseType: ResponseType;
   state?: string;
   verifier?: string;
   query?: StringDict<string>;
+  prompt?: 'none' | 'login' | 'consent';
 }
 
 /**
--- a/packages/javascript-sdk/src/oauth2-client/interfaces.ts
+++ b/packages/javascript-sdk/src/oauth2-client/interfaces.ts
@@ -11,6 +11,7 @@
 import { StringDict } from '../shared/interfaces';
 import { ConfigOptions } from '../config';
 import { ResponseType } from './enums';
+import { GetTokensOptions } from '../token-manager';
 
 /**
  * Tokens returned after successful authentication.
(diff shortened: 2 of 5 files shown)


packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -158,6 +162,12 @@ abstract class OAuth2Client {
     const responseBody = await this.getBody<unknown>(response);
 
     if (response.status !== 200) {
+      const error_description_query_param =
+        new URL(response.url).searchParams.get('error_description') !== undefined;
+
+      if (error_description_query_param) {
+        throw new Error(allowedErrors.PromptError);
+      }
```
ryanbas21, line 170: honestly unsure if this was really a great way to handle this. need to look into this more

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -38,6 +38,9 @@ const allowedErrors = {
   NetworkError: 'NetworkError when attempting to fetch resource.',
   // Webkit browser error
   CORSError: 'Cross-origin redirection',
+
+  //Prompt error
+  PromptError: 'User is not authenticated or session cookie not sent to the server',
```
cerebrl, line 43: I think the `error` parameter may be better than the `error_description` parameter. Rather than this long sentence, you'd look for this: `interaction_required`.

cerebrl: Actually, looking at this now, the most minimal code change would be to use both the `error` and `error_description` as defined by this `parseError` method: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/oauth2-client/index.ts#L293.

packages/javascript-sdk/src/oauth2-client/index.ts
cerebrl: I don't think this is needed. The `parseError` method used just a few lines down should already grab the error and throw with the right message.

packages/javascript-sdk/src/oauth2-client/interfaces.ts
```diff
@@ -35,7 +36,7 @@ interface AccessTokenResponse {
 /**
  * Options used when requesting the authorization URL.
  */
-interface GetAuthorizationUrlOptions extends ConfigOptions {
+interface GetAuthorizationUrlOptions extends GetTokensOptions {
```
cerebrl, line 39: Was this intentional?

cerebrl (commented):
Just a few comments.

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -50,6 +53,7 @@ abstract class OAuth2Client {
     const requestParams: StringDict<string | undefined> = {
       ...options.query,
       client_id: clientId,
+      prompt: options.login ? options.login : 'none',
```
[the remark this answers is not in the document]
ryanbas21: Good catch. Honestly, i think this part of the jira description made me think options.login made sense, but seeing the docs here i know its now wrong.

```
When creating the authorization URL for the hidden iframe, use `prompt` of "none"
For Central Login, the prompt should be removed or the use of "login" replaces "none"
```

I thought "login" would replace 'none'

packages/javascript-sdk/src/oauth2-client/interfaces.ts
cerebrl: This doesn't feel right to me as the hierarchy of this is in reverse. The `TokenManager` class, on which `getTokens` is found, internally uses the `OAuthClient` class. So, if one inherited from the other, I would think `GetTokensOptions` would extend `GetAuthorizationUrlOptions`, not the other way around, yeah? But, ultimately, I don't think either extend the other. For example, `forceRenew` and `login` are not properties of `GetAuthorizationUrlOptions` as the `createAuthorizeUrl` doesn't use them.

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -38,6 +38,9 @@ const allowedErrors = {
   NetworkError: 'NetworkError when attempting to fetch resource.',
   // Webkit browser error
   CORSError: 'Cross-origin redirection',
+
+  // prompt=none errors
+  InteractionNotAllowed: 'The request requires some interaction that is not allowed.',
```
cerebrl, line 43: I'm a bit confused here. Being that the `parseError` method concatenates the `error` and `error_description`, I would think this string would start with `interaction_required: `: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/oauth2-client/index.ts#L296. Am I misunderstanding what happens here?

cerebrl (changes requested):

packages/javascript-sdk/src/oauth2-client/index.ts
ryanbas21: Screen Shot 2023-02-15 at 2 52 49 PM

packages/javascript-sdk/src/oauth2-client/interfaces.ts
```diff
@@ -35,11 +36,12 @@ interface AccessTokenResponse {
...
   responseType: ResponseType;
   state?: string;
   verifier?: string;
   query?: StringDict<string>;
+  prompt?: 'none' | 'redirect' | 'consent';
```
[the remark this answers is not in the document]
ryanbas21: d'oh. yes.

packages/javascript-sdk/src/oauth2-client/index.ts
cerebrl: Ah, okay, you're right. Sorry for the confusion.

packages/javascript-sdk/src/token-manager/index.ts
```diff
@@ -118,9 +118,12 @@ abstract class TokenManager {
      */
     const verifier = PKCE.createVerifier();
     const state = PKCE.createState();
-    const authorizeUrlOptions = { ...options, responseType: ResponseType.Code, state, verifier };
-    const authorizeUrl = await OAuth2Client.createAuthorizeUrl(authorizeUrlOptions);
-
+    const authorizeUrlOptions = {
```
ryanbas21, line 121: this got formatted /shrug

packages/javascript-sdk/src/token-manager/index.ts
```diff
@@ -165,6 +169,9 @@ abstract class TokenManager {
 
       // Since `login` is configured for "redirect", store authorize values and redirect
       window.sessionStorage.setItem(clientId as string, JSON.stringify(authorizeUrlOptions));
+
+      const authorizeUrl = await OAuth2Client.createAuthorizeUrl(authorizeUrlOptions);
```
ryanbas21, line 173: i moved this from above, not needed but its closer to where its used now

cerebrl (approved):

packages/javascript-sdk/src/oauth2-client/interfaces.ts
ryanbas21: we resolved this on slack. its right and wrong and i've created a ticket to complete a fix for this.

packages/javascript-sdk/src/oauth2-client/interfaces.ts
```diff
@@ -40,6 +40,7 @@ interface GetAuthorizationUrlOptions extends ConfigOptions {
   state?: string;
   verifier?: string;
   query?: StringDict<string>;
+  prompt?: 'none' | 'login' | 'consent';
```
daveadams56, line 43: My understanding is that the requirement in the ticket was for `none` to mean that `?prompt=none` is passed in the `authorize` call, `login` and would be used for centralized login. Maybe a silly question, but what does `consent` mean here - what is the consequence of specifying this option?

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -82,7 +85,8 @@ abstract class OAuth2Client {
    * New Name: getAuthCodeByIframe
    */
   public static async getAuthCodeByIframe(options: GetAuthorizationUrlOptions): Promise<string> {
-    const url = await this.createAuthorizeUrl(options);
+    const url = await this.createAuthorizeUrl({ ...options, prompt: 'none' });
```
daveadams56, line 88: Given that we are setting prompt=none as a default, does it make sense to make this something the developer can specify? As I understand it, setting `prompt: 'login'` overrides this but if nothing is set then it's `prompt: 'none'`. Maybe I'm overthinking this but what do you think?

daveadams56 (commented):
I see this was merged before I hit 'submit' on my review so apologies for not getting feedback to you before this, however I have a couple of questions around the implementation. Could both be things that I missed, but it would be good to discuss these here or in slack when you get time. Thanks.

packages/javascript-sdk/src/oauth2-client/interfaces.ts
cerebrl: This is a good point. These three values for the `/authorize` endpoint are valid, but we are likely unable to support the `'consent'` option. Should remove this option as to not confuse the dev?

packages/javascript-sdk/src/oauth2-client/index.ts
cerebrl: Good question. This is how I look at it: since this is within the hidden iframe, I don't believe there is ever a reason to NOT use `prompt=none` as any prompt, aka UI, would be hidden and therefore unusable.

packages/javascript-sdk/src/oauth2-client/interfaces.ts
daveadams56: I think this is misleading if it doesn't actually do anything or isn't compatible in the context of the SDK and should be removed. I'd worry that it might inadvertently trick developers into thinking some action should occur if they specify consent, for example showing a consent collection page.

packages/javascript-sdk/src/oauth2-client/index.ts
daveadams56: Yes. Given this and my point about the value `consent` further up, perhaps we should rethink this. Maybe we should obscure this entirely from the user and always pass `prompt=none` unless the developer specifies they want to do centralized login using `login: 'redirect'`? That way we avoid introducing a "thing" that the developer has to think about.

PR #330, opened 2023-02-16
fix: correct code / type in  getTokens call

ryanbas21:
# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2375

# Description
the options call is a GetTokens but the options we spread in should only be a ConfigOptions. so we can remove the types from the call that are not config options and pass them into authorization url creation

# Type of Change

# How Has This Been Tested?

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - master
+  workflow_dispatch:
 env:
   NX_CLOUD_DISTRIBUTED_EXECUTION: true
 jobs:
@@ -59,7 +60,6 @@ jobs:
       - run: npm ci
         if: steps.cache.outputs.cache-hit != 'true'
         env:
-          # PLAYWRIGHT_BROWSERS_PATH: 0
           PLAYWRIGHT_SKIP_BROWSERS_DOWNLOAD: 1
       - run: npm install @nrwl/nx-cloud
       - name: Start Nx Agent ${{ matrix.agent }}
@@ -75,18 +75,22 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ssh-key: ${{ secrets.SSH_PUBLIC_KEY }}
           persist-credentials: false
-      - uses: webfactory/ssh-agent@v0.5.4
-        with:
-          ssh-private-key: ${{ secrets.SSH_KEY }}
-      - name: Set git config username and email
-        run: git config user.email "ryan.basmajian@forgerock.com" && git config user.name "ryanbas21"
-        shell: bash
       - uses: actions/setup-node@v3
         id: cache
         with:
           node-version: '16.x'
           scope: '@forgerock'
+      - name: Import GPG
+        uses: crazy-max/ghaction-import-gpg@v5.2.0
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+          git_committer_email: 'ryan.basmajian@forgerock.com'
+          git_push_gpgsign: true
+          git_config_global: true
       - run: npm ci
       - run: npm install @nrwl/nx-cloud
       - run: npm run build
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@commitlint/config-conventional": "^17.4.2",
     "@commitlint/config-nx-scopes": "^17.4.2",
     "@commitlint/prompt": "^16.2.3",
-    "@jscutlery/semver": "2.29.3",
+    "@jscutlery/semver": "2.30.1",
     "@nrwl/angular": "15.6.3",
     "@nrwl/cli": "15.6.3",
     "@nrwl/eslint-plugin-nx": "15.6.3",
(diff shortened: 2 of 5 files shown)


.github/workflows/merge.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - master
+  workflow_dispatch:
```
ryanbas21, line 7: allows this to be run wihtout a pr (mostly for me sanity checking or re reunning things or if we want to cut a release without a pr)

.github/workflows/merge.yml
```diff
@@ -75,18 +75,23 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ssh-key: ${{ secrets.SSH_PUBLIC_KEY }}
```
ryanbas21, line 78: authenticate clone the repo

.github/workflows/merge.yml
```diff
@@ -75,18 +75,23 @@ jobs:
...
-      - name: Set git config username and email
-        run: git config user.email "ryan.basmajian@forgerock.com" && git config user.name "ryanbas21"
         shell: bash
       - uses: actions/setup-node@v3
         id: cache
         with:
           node-version: '16.x'
           scope: '@forgerock'
+      - name: Import GPG
+        uses: crazy-max/ghaction-import-gpg@v5.2.0
```
ryanbas21, line 87: sign the commits. its locked to v5.2.0 as per our old discussion

.github/workflows/merge.yml
```diff
@@ -75,18 +75,23 @@ jobs:
...
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
```
ryanbas21, line 89: this is my gpg private key

package.json
```diff
@@ -92,7 +92,7 @@
     "ts-loader": "^8.0.17",
     "ts-node": "10.9.1",
     "typedoc": "^0.17.7",
-    "typescript": "4.8.4",
+    "typescript": "4.9.5",
```
ryanbas21, line 95: satisfies operator and newer ts stuff

packages/javascript-sdk/src/token-manager/index.ts
```diff
@@ -118,8 +118,9 @@ abstract class TokenManager {
      */
     const verifier = PKCE.createVerifier();
     const state = PKCE.createState();
+    const { forceRenew, login, query, ...config } = options;
```
ryanbas21, line 121: unsure if theres a better way to remove the lint warnings on unused variables but this felt like the cleanest option.

daveadams56: Ok, think I understand this now after reading through for a little bit. Not sure how we can remove the lint warnings but will have a think. What you're doing makes sense though.

daveadams56 (commented):
Initial look - seems good, I had to think a bit to understand what we are doing here but it makes sense to me now. I'll have a quick think about linting warnings if I get a chance.

packages/javascript-sdk/src/token-manager/index.ts
cerebrl: I think we need to pass `query` into the `authorizeUrlOrptions` as well, yeah? It's used here: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/oauth2-client/index.ts#L53.

daveadams56: Yes I think you're right, good spot. Yeah it looks like `GetTokensOptions` defines query on top of `ConfigOptions`.

.github/workflows/merge.yml
daveadams56: I had a look at the file separately as it looked odd - l shouldn't the `workflow_dispatch` key have a value, or some child keys indented below it? At first glance I assumed env: was a child of this but now seeing it without this review block in the way I don't quite understand how this is valid.

package.json
daveadams56: Just wondering we have something for this project like dependabot that we use (whether manual or automated), to identify which of these dependencies need updating and when?

ryanbas21: we have dependabot but not configured or setup really like i did in the other repo

packages/javascript-sdk/project.json
```diff
@@ -81,7 +81,7 @@
           "commitMessageFormat": "release: ${version} [skip ci]",
           "remote": "origin",
           "push": true,
-          "postTargets": ["javascript-sdk:deploy:beta"],
+          "postTargets": ["javascript-sdk:deploy:beta", "javascript-sdk:github:beta"],
```
daveadams56, line 84: Are you able to explain what we're adding here, with `javascript-sdk:github:beta` and `javascript-sdk:github`? It's not obvious to me and would like to know more. Thanks.

ryanbas21: its a post target, so after this runs, it will run that nx target.

.github/workflows/merge.yml
ryanbas21: So we don't really pass any inputs into this workflow, so i don't think we need anything.  Truthfully im not entirely sure we need this to begin with, i was adding it as a way of running CI without needing a PR up. so if you just need to sanity check things or something odd fails and you want to re run the branch

cerebrl (approved):

PR #350, opened 2023-06-12
feat(token-vault): implements the token vault plugin feature for sdk

cerebrl:
# JIRA Ticket

- [Token Vault Proxy – SDKS-2487](https://bugster.forgerock.org/jira/browse/SDKS-2487)
- [Token Vault Interceptor – SDKS-2486](https://bugster.forgerock.org/jira/browse/SDKS-2486)
- [Token Vault Client (Plugin) – SDKS-2488](https://bugster.forgerock.org/jira/browse/SDKS-2488)

# Description

[NOTE: to simplify reviewing the changes specifically for the Token Vault feature itself, and NOT for the infrastructure within the repo to support the feature, there are two PRs:

1. The infrastructure setup: https://github.com/ForgeRock/forgerock-javascript-sdk/pull/348
2. This PR that builds on top of that setup and adds the feature completion to Token Vault

That means this PR has a dependency on the previous PR to be merged.]

This is the design document for the development of the Token Vault (aka "Origin Isolation") Plugin & Proxy for the JavaScript SDK. This uses a Service Worker for intercepting `fetch` requests and rerouting them through an [iframe with an alternative origin (Origin Isolation)](https://github.com/ForgeRock/appAuthHelper/blob/master/origin_isolation.md). By leveraging an iframe on an alternative origin (importantly, the same parent domain but different subdomain), we are able to store the tokens within the alt. origin app inside the `iframe` using the standard Web Storage API, while also keeping the tokens completely inaccessible to the main application due to it being on a different origin. This is the Token Vault.

There are 3 main entities/processes involved in the system:

1. **Main App**: Standard JavaScript SPA (running on `https://app.example.com`)
4. **Service Worker**: registered on Main App (running on `https://app.example.com`)
5. `iframe` App: injected into Main App (running on `https://proxy.example.com`) aka the "Token Vault Proxy"

It's worth noting that this is heavily inspired by [AppAuthHelper](https://github.com/ForgeRock/appAuthHelper) and the [patent filed from Jake on behalf of ForgeRock](https://patents.justia.com/patent/20210084038). This project is intended to provide a solution that integrates the functionality of AppAuthHelper within an application that uses the [ForgeRock JavaScript SDK](https://github.com/ForgeRock/forgerock-javascript-sdk).

To accomplish this system design, there are 3 npm modules involved:

1. **JavaScript SDK**: provides existing APIs for interacting with AM as well as the token store API that the Token Vault plugin will use
2. **Token Vault Plugin for SDK**: provides the Service Worker for intercepting requests and the Post messaging system to communicate with the Identity Proxy module
3. **Token Vault Proxy** (`iframe`): provides the token store, HTTP request forwarding with automatic token injection and token renewal with Refresh Token

# Type of Change

- [x] This change requires a documentation update

# Code Review Note

When reviewing this PR, the changes found in these directories are the most important as they contain the source code:

- `/packages/token-vault/`
- `/shared/network/`
- `/shared/workers`

This is the e2e test app and the code that represents _how_ a developer to use this:

- `/e2e/token-vault-app/`
- `/e2e/token-vault-proxy/`

This will contain the E2E automation:

- `/e2e/token-vault-suites/`

# How Has This Been Tested?

**Unit tests**: The mostly functional utility functions are unit tested via Jest. Much of the code within this new feature is almost entirely made of side-effects, so these tests don't cover the a lot of the functionality.

**E2E tests**: The infrastructure is setup for E2E testing, but there are no tests yet, but this will be developed over the next few days.

**Manual testing**: There's a test app configured in `/e2e/token-vault-app/`. Running these two commands in separate terminal windows to run both apps: 1) `npx nx run token-vault-app:serve`, 2) `npx nx run token-vault-proxy:serve`. The buttons rendered to page result in console messages, and you have to validate what you see in the dev-tools.

# Definition of Done

Check all that apply

- [x] Coded to standards.
- [ ] Code peer-reviewed.
- [x] Unit tests are written.
- [ ] Integration tests are written (likely won't use integration tests for this feature)
- [ ] e2e tests are written (coming soon)
- [x] Functional spec is written/updated
- [x] contains example code snippets.
- [ ] Change log updated.
- [x] Documentation story is created and tracked.
- [x] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


e2e/token-vault-app/src/main.ts
```diff
@@ -1,40 +1,50 @@
...
-// Register the proxy
+// Register the Token Vault Proxy
 const proxy = await register.proxy(
   document.getElementById('token-vault') as HTMLElement
 );
 
-// Register the token store replacement
-const storeReplacement = register.store();
+// Register the Token Vault Store
+const tokenStore = register.store();
```
cerebrl, line 35: Token Vault Client initialization and registration of Interceptor (service worker), Proxy (iframe) and Store. The values returned to `interceptor` and `proxy` are optional and used only if the dev will want to "destroy" the iframe or unregister the service worker.

e2e/token-vault-proxy/src/main.ts
```diff
@@ -3,16 +3,15 @@ import { proxy } from '@forgerock/token-vault';
 // Initialize the token vault proxy
 proxy({
```
cerebrl, line 4: This is the proxy (iframe) initialization call, passing in the configuration to setup it's connection with the main app and the ForgeRock platform.

e2e/token-vault-app/src/main.ts
```diff
@@ -1,40 +1,50 @@
...
   Config,
   FRUser,
   TokenManager,
   UserManager,
 } from '@forgerock/javascript-sdk';
-// Initialize the token vault client
+import { client } from '@forgerock/token-vault';
+
+/** ****************************************************
+ * START: CONFIGURE THE TOKEN VAULT
```
cerebrl, line 10: The vast majority of this code has been liberally comments like this. Please read the code comments that they try to be as informative as possible.

packages/token-vault/src/lib/client.ts
```diff
@@ -9,30 +10,51 @@ interface ClientConfig extends ClientConfigInit {
...
+ *     scope: '/',
+ *   },
+ *   proxy: {
+ *     origin: 'http://proxy.example.com',
+ *     url: 'http://proxy.example.com/path',
+ *   },
+ * });
  */
-export function client(config: ClientConfig) {
+export function client(config: ClientConfig): ClientInit {
```
cerebrl, line 37: This is the Token Vault Client initialization code you see in the `main.ts` app within the `e2e/token-vault-app/` directory. This returns the registration object to register all the entities in the system.

packages/token-vault/src/lib/client.ts
```diff
@@ -9,30 +10,51 @@ interface ClientConfig extends ClientConfigInit {
...
 }
 
-/**
- * @function client - Initialize the token vault client
- *
- * @param {Object} config
- *
- * @returns {undefined}
+/** ****************************************************************
+ * @function client - Initialize the Token Vault Client
```
cerebrl, line 14: Again, pay close attention to the code comments throughout.

packages/token-vault/src/lib/client.ts
```diff
@@ -45,19 +67,25 @@ export function client(config: ClientConfig) {
...
     },
+
+    /** ****************************************************
+     * @method proxy - Register the Token Vault Proxy
+     * @param {HTMLElement} target - The target element to append the proxy iframe to
+     * @param {BaseConfig['proxy']} options - The configuration object for the proxy
+     * @returns {void}
+     * @example register.proxy(document.getElementById('token-vault'));
+     */
     proxy: function (target: HTMLElement, options?: BaseConfig) {
```
cerebrl, line 78: This dynamically adds the iframe to the DOM to register the Token Vault Proxy.

packages/token-vault/src/lib/client.ts
```diff
@@ -88,18 +116,35 @@ export function client(config: ClientConfig) {
...
           // Tokens are not retrievable from the iframe
           return Promise.resolve(null as unknown as Tokens);
         },
-        has() {
+
+        /**
+         * @method has - Check if tokens exist in the Token Vault
+         * @returns {Promise<{ hasTokens: boolean }>} - True if tokens exist, false if not
+         */
+        has(): Promise<{ hasTokens: boolean }> {
```
cerebrl, line 147: You will notice the two additional methods for the main app to directly ask the Proxy a question or request an action. Notice that none return tokens.

This one is just asking the Proxy if it has tokens, and it returns `{ hasTokens: boolean }`.

packages/token-vault/src/lib/client.ts
```diff
@@ -113,7 +158,12 @@ export function client(config: ClientConfig) {
             };
           });
         },
-        refresh() {
+
+        /**
+         * @method refresh - Refresh tokens in the Token Vault
+         * @returns {Promise<{ refreshTokens: boolean }>}
+         */
+        refresh(): Promise<{ refreshTokens: boolean }> {
```
cerebrl, line 166: This allows the main app to directly ask the Proxy to refresh the Access Token. It does not return the tokens.

packages/token-vault/src/lib/exchange.ts
cerebrl, line 1: Unused file that was deleted.

packages/token-vault/src/lib/proxy.ts
cerebrl, line 1: This is the longest file in the feature. I'm thinking we could convert this into a "router" of types. Maybe like Express, Sinatra or some other route driven, lightweight framework. It would route the event to the appropriate event handler and, if event type is fetch, then it would route the URL to the appropriate handler. But, until then, we are just using conditionals to filter through the events.

Read the code comments!

packages/token-vault/src/lib/proxy.ts
```diff
@@ -1,60 +1,101 @@
...
    */
-  const fetchEventName = config?.events?.fetch || 'FETCH_RESOURCE';
-  const hasTokenEventName = config?.events?.has || 'HAS_TOKENS';
-  const refreshTokenEventName = config?.events?.refresh || 'REFRESH_TOKENS';
-  const removeTokenEventName = config?.events?.remove || 'REMOVE_TOKENS';
+  const eventsObj = eventDefaults.reduce((acc, event) => {
+    const eventName = config.events?.[event.type] || event.name;
+    acc[event.type] = eventName;
+    return acc;
+  }, {} as EventsConfig);
```
cerebrl, line 55: This is kind of a fancy way to generate the event objects for the conditionals below with the capability to customize them by the dev, if they wanted.

packages/token-vault/src/lib/proxy.ts
```diff
@@ -1,60 +1,101 @@
...
   /**
    * Generate AM URLs
    */
-  const urls = generateUrls(config?.forgerock);
+  const urls = generateAmUrls(config?.forgerock);
 
   /**
    * Create the proxy iframe
    */
   window.addEventListener('message', async (event) => {
```
cerebrl, line 65: This is the one and only event listener that captures all "events" emitted from the main app.

shared/workers/src/lib/interceptor.ts
```diff
@@ -0,0 +1,100 @@
...
+          proxyChannel.port1.onmessage = (messageEvent) => {
+            console.log(`Returning ${url}`);
+            const response = messageEvent?.data || {};
+
+            /**
+             * Create a new response from the response body and headers
+             * The body, first argument, needs to be converted back to string
+             */
+            resolve(
+              new Response(JSON.stringify(response?.body), {
```
cerebrl, line 89: It's important to note that all request and responses are cloned and serialized in order to send them back and forth between the entities. The contract between the Interceptor and the Proxy is that native requests get converted into a request-like object; the proxy converts the request-like object back into a true request. The same, happens on the Proxy side with responses.

packages/token-vault/src/lib/client.ts
```diff
@@ -88,18 +116,35 @@ export function client(config: ClientConfig) {
...
           return Promise.resolve(null as unknown as Tokens);
         },
-        has() {
+
+        /**
+         * @method has - Check if tokens exist in the Token Vault
+         * @returns {Promise<{ hasTokens: boolean }>} - True if tokens exist, false if not
+         */
+        has(): Promise<{ hasTokens: boolean }> {
           const proxyChannel = new MessageChannel();
```
cerebrl, line 148: There could be a performance consideration with creating a message channel for each of these calls. Mainly if the calls are done in a long loop and these message channels are rapidly generated faster than they can be cleaned up. It may take the Garbage Collector a while to clean them up.

The reason for doing this, though, is to ensure reference identity for the request-response pairing. Creating a single message channel and trying to reuse it, you could have race conditions resulting in responses that are mismatched with requests.

shared/network/README.md
```diff
@@ -0,0 +1,11 @@
+# shared-network
```
ryanbas21, line 1: I think we should take advantage of readmes in shared libraries to explain the purposes and maybe document pieces that otherwise would be lost in code.

shared/types/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@forgerock/shared-types",
+  "name": "@shared/types",
```
ryanbas21, line 2: hmm was this me or you? I think we should not do `@` in the name of a packge.json because to `npm` that infers its a scoped package (while this isn't an npm package, i just think it may be a good idea). Thoughts?

shared/types/src/lib/config.types.ts
```diff
@@ -0,0 +1,99 @@
+/** ****************************************************************
+ * This contains a few types pulled from the JavaScript SDK
+ * TODO: Refactor the SDK to use these shared types instead of the internal ones
+ */
+import type { ConfigOptions } from '@forgerock/javascript-sdk';
+
+/** ****************************************************************
+ * NOTE: Pulled out of the SDK's /config/enums.ts file
+ */
+export enum ActionTypes {
```
ryanbas21, line 10: I'd favor a union over enum. its a slightly debated topic but most people say go with unions now.

ryanbas21: we can always refactor this when we do the `NOTE`

ryanbas21 (changes requested):
initial review, mostly just familiarizing myself with parts. My initial feelings about the api are good. I think there are a reasonable amount of sensible defaults at the config level and I think the api itself is straightforward. 

I have to revisit the "replacement" of sdk methods we pass in like "get" to understand why we are doing that more.

e2e/token-vault-proxy/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@forgerock/token-vault-proxy",
+  "description": "ForgeRock Token Vault Test Proxy",
+  "private": true,
+  "devDependencies": {
+    "@vitejs/plugin-basic-ssl": "^1.0.1"
```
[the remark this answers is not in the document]
cerebrl: No, I removed it locally, and will push it with the next batch of changes.

e2e/token-vault-suites/src/app.test.ts
```diff
@@ -0,0 +1,36 @@
+import { expect, test } from '@playwright/test';
+import { asyncEvents } from './utils/async-events';
```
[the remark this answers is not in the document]
cerebrl: Yeah, we can do that. We can discuss this when we have our conversation later today.

packages/token-vault/src/lib/client.ts
```diff
@@ -88,18 +116,35 @@ export function client(config: ClientConfig) {
...
 
       return {
-        get() {
+        /**
+         * @method get - A noop method that replaces the default get method
+         * @returns {Promise<null>}
+         */
+        get(clientId: string): Promise<Tokens> {
           // Tokens are not retrievable from the iframe
           return Promise.resolve(null as unknown as Tokens);
```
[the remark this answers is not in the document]
cerebrl: To be honest, I don't remember. I can remove it and see if I can recreate the issue that led to it.

shared/network/README.md
cerebrl: Yeah, I agree. I'll try to find some time and add some content here.

shared/types/package.json
cerebrl: Yeah, I'm okay with using a different symbol, or were you wanting no symbol at all? I kind of like the symbols as it makes it easier to know from scanning when something is a "special module" versus just a file within system.

shared/types/src/lib/config.types.ts
cerebrl: Yeah, I'm with you about enums: I'm not a fan either. This was just copied from the original SDK code, and I didn't want to refactor it at the time. Once we get our vision with our next steps concerning "shared" code, we can address it then.

packages/token-vault/src/lib/client.ts
```diff
@@ -45,19 +67,25 @@ export function client(config: ClientConfig) {
 
       return await registerServiceWorker();
```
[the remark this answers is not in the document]
ryanbas21: It's just semantics I suppose, but yes.

I think we can have an `await` on its own line if we are managing the failure of the promise here in this function with a try catch as the `return <my-promise>` will force the error to bubble up to the next layer in the stack, where as breaking it into two lines would have it error within this function.

Since we are not managing any failure here we could just `return registerServiceWorker()` and it should behave the same.

e2e/token-vault-suites/.eslintrc.json
```diff
@@ -4,11 +4,15 @@
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
-      "rules": {}
+      "rules": {
+        "jest/no-done-callback": "off"
```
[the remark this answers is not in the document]
ryanbas21: let's remove it, since we never use `done` in tests anyways

shared/types/package.json
ryanbas21: lets add this to our discussion

e2e/token-vault-proxy/index.html
```diff
@@ -0,0 +1,16 @@
...
+    <meta charset="utf-8" />
+    <title>OAuth Callback Handler</title>
+    <base href="/" />
+
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="icon" href="favicon.ico" />
+    <link rel="stylesheet" href="/src/styles.scss" />
+  </head>
+  <body>
+    <forgerock-root></forgerock-root>
```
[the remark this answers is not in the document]
ryanbas21: This is because the `web` generator technically makes a web component library. so we could "write web components" which that would be.

ryanbas21 (approved):
@lsanches knows about the mend failure. we are ignoring it for now.

PR #368, opened 2023-07-13
fix(javascript-sdk): get-tokens-default-destructure

ryanbas21:
# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2570

# Description

TokenManager.get tokens when nothing is past in can break in module systems that are not downleveled to supply the default. this change adds a fix where we supply a default to the destructuring so that this doesn't occur

# Type of Change

# How Has This Been Tested?

verified via unit test and i copied the dist from the javascript-sdk build and put it in the app where the bug was discovered with no downleveling and it worked.

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread

### DIFF
--- a/packages/javascript-sdk/src/token-manager/index.ts
+++ b/packages/javascript-sdk/src/token-manager/index.ts
@@ -118,9 +118,12 @@ abstract class TokenManager {
     const verifier = PKCE.createVerifier();
     const state = PKCE.createState();
 
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    const { forceRenew, login, ...config } = options;
+    // so to not change the type of the above function
+    // we assign it here if its undefined or null.
+
+    const config = Object.assign({}, options);
+    delete config.forceRenew;
+    delete config.login;
 
     const authorizeUrlOptions = {
       ...config,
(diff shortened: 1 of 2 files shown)


packages/javascript-sdk/src/token-manager/index.ts
```diff
@@ -118,9 +118,16 @@ abstract class TokenManager {
...
+    // so to not change the type of the above function
+    // we assign it here if its undefined or null.
+
+    if (!options) options = {};
+
+    // in the scenario above where an options object is undefined or null
+    // and we try to destructure a value from an empty object
+    // we assign defaults
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { forceRenew = false, login = true, ...config } = options;
```
cerebrl, line 130: I'm okay with this, but if we have an alternative that doesn't require the disabling of rules and other stuff, I think I'd prefer that instead. Is this any better:

```js
const config = Object.assign({}, options);
delete config.forceRenew
delete config.login
```

The above doesn't need the condition, the comments or the rule disabling. I think it's quite self-explanatory. Another plus is we aren't actually using the `forceRenew` and `login` that we are pulling out (hence the rule disabling), we're just using them to get to the rest (no pun intended).

cerebrl (approved):
My comment is a non-blocker, but I think I'd prefer an alternative that requires less "stuff". Let me what you think.

PR #380, opened 2023-07-31
[Sample App] Create a social login sample app in SDK

juanManuel05:
# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2538

# Description
[Sample App] Create a social login sample app in SDK
# Type of Change

# How Has This Been Tested?

**1) Local Log-in**

**STEPS TO TEST IT:**

- if not user, create a new user
- Click on Sign in
- Enter password and user, hit enter or click on Submit
- The page will check your credentials and, if successful, it will finish the log in flow.
- Go to Todo section
- create a new todo, should add it with no error
- Log out and log in, the previously todo should now appear on the list.

**2) Social-media log-in**

 **Pre- requisites:** The social identity provider service must be created and configured for the tenant. For this case, google social provider has been added as social identity provider service for the tenant only, via Access Management. 
**Important:** It may be necessary to clear site data before/after every log-in attempt
Next the configuration recommended for testing purposes:

AM_URL=https://openam-sdks.forgeblocks.com/am/
APP_URL=https://angular.example.com:8443
API_URL=http://localhost:9443
DEBUGGER_OFF=true
JOURNEY_LOGIN=google-login-web
JOURNEY_REGISTER=Registration
REALM_PATH=alpha
WEB_OAUTH_CLIENT=WebOAuthClient
PORT=9443
REALM_PATH=alpha
REST_OAUTH_SECRET=mysecret
REST_OAUTH_CLIENT=RestOAuthClient

**STEPS TO TEST IT:**

- Click on Sign in
- The page will provide as many options as  social identity provider services the tenants has configured 
- Click on Sign in with google
- Enter google credentials
- Enter ForgeRock credentials
- A notification will be sent to the Forge Rock authenticator App
- Approve notification
- The page will check your credentials and, if successful, it will finish the log in flow.
- Go to Todo section
- create a new todo, should add it with no error
- Log out and log in, the previously todo should now appear on the list.
[ForgeRock_ A Sample Angular (Web) Todo App.webm](https://github.com/ForgeRock/forgerock-javascript-sdk/assets/18680830/91f3619a-abaf-4973-8994-443ef753daec)

Next how the IdPs selection looks like

Screenshot 2023-08-08 at 15 25 44

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


ryanbas21 (changes requested):
Remember, if you are adding environment variables to please update the env.example file, as well as the angular environment file (unless theres a way we can avoid this horrible file).

Thanks

samples/angular-todo/src/environments/environment.ts
```diff
@@ -1,8 +1,8 @@
 export const environment = {
-  AM_URL: 'https://openam-crbrl-01.forgeblocks.com/am',
+  AM_URL: 'https://openam-sdks.forgeblocks.com/am/',
   REALM_PATH: 'alpha',
   WEB_OAUTH_CLIENT: 'WebOAuthClient',
-  JOURNEY_LOGIN: 'Login',
+  JOURNEY_LOGIN: 'google-login-web',
```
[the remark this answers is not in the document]
juanManuel05: Upon this one, the **google-login-web** journey shows both, a _username-and-password log-in_ as well as a _social-login_ option.
image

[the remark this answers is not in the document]
ryanbas21: so in the case of this journey, if the social login variable is provided, we use that. if it isn't we would fall back to the standard Login tree, if thats not passed in, then the app just wouldnt work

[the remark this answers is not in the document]
ryanbas21: Thinking about this again, this is fine, not sure where my head was going the other day.

samples/angular-todo/src/app/app.component.ts
```diff
@@ -65,7 +65,7 @@ export class AppComponent implements OnInit {
 
     Config.set({
       clientId: environment.WEB_OAUTH_CLIENT,
-      redirectUri: `${window.location.origin}/callback`,
```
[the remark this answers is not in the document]
juanManuel05: i'm asking about this as if i check the routing file, the path 'callback' does not exist and therefore throws an error
image
image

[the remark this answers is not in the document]
cerebrl: I see what Juan is trying to fix, and I think we should address it. The question is just how we fix it. Currently, both the React app and the Angular app use a redirect URL that actually doesn't point to something that's properly handled. The Angular app may just be more sensitive to this, as it results in an error seen in Juan's screenshot.

Ultimately, I don't feel this redirect URL should result in the Angular app itself, or really anything that uses JS and CSS, being loaded and rendered. Maybe we could point to a static file, rather than a "route"? So, instead of `${window.location.origin}/home` or `${window.location.origin}/callback`, we could use `${window.location.origin}/callback.html`. The server should recognize that as a static resource, rather than a route in the main app, which results in quicker load times and less resource usage.

Something along the lines of a plain old HTML file that just renders `<!DOCTYPE html><html><p>OK</p></html>`. That way, the page loads and the SDK can quickly read the final URL of the resolved request. Apps with embedded login will never actually see the rendered page, so this makes sense to me.

Thoughts?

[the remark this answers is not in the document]
juanManuel05: i fixed by adding a new path into the routing file, when '/callback' it loads the HomeComponent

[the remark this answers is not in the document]
ryanbas21: @cerebrl yeah, it was pointing at a callback.html in the assets directory but it seems angular router doesn't like to not know about possible routes. so Juan was able to correct it by assigning a router so angular can say it knows about the route. That seems to fix the angular issue we had previously.

So if its fixed, i don't think webpack dev server was rewriting the params, it was angular rejecting the route because it was not aware of it.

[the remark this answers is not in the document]
cerebrl: This is a solution, but I'm not sure it's one that we should use long-term. We don't have to address it in this PR, but here's my reasoning:

The request to `/callback` due to the redirect from `/authorize` just has to resolve when it's a part of Embedded Login. It doesn't have to render anything as everything within the iframe will be hidden when using Embedded Login. Due to this, all the resources the full Angular app is dependent on has to be loaded onto the page in order to call the `onload` event found here: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/oauth2-client/index.ts#L110. This is required for the token request to be continued.

Regarding the `load` event ([from MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event)):

> The load event is fired when the whole page has loaded, including all dependent resources such as stylesheets, scripts, iframes, and images.

To load up an entire Angular application just to resolve a URL that will never be rendered is ... well, a lot. To keep this app as "real world" as we can, I would suggest we do what we would recommend to our customers, which is have `/callback`, or whatever we call it, refer to a static asset, rather than an full JS application.

Now, if this was Centralized Login, it would be a different story as the redirect URL is used in a full-page redirect, rather than a hidden iframe. So, context matters.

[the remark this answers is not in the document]
ryanbas21: Hmm, i could have sworn this was the behavior at one point. I don't disagree, so lets see if we can do that Juan.

Add a basic html file (i feel we've done this before) called callback.html. We will need to serve that file as a static asset. 

I think you will add it to the `assets` key in project.json under serve, so it serves callback.html (i'd keep it at the same level of the index.html).

See if we can serve this file without an angular route specifically.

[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -194,4 +240,12 @@ export class FormComponent implements OnInit {
       }
     }
   }
+
+  isIdentityProviderLogin(callback: SelectIdPCallback): boolean {
+    return callback
```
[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/environments/environment.ts
```diff
@@ -1,8 +1,8 @@
 export const environment = {
-  AM_URL: 'https://openam-crbrl-01.forgeblocks.com/am',
+  AM_URL: 'https://openam-sdks.forgeblocks.com/am/',
```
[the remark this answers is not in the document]
juanManuel05: not too much :/

samples/angular-todo/src/app/features/journey/identityProvider/identity-provider.component.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.identityProviders = this.callback
+      .getProviders()
+      .filter((provider) => provider.provider !== 'localAuthentication');
+  }
+
+  onSetProvider(provider: string): void {
+    this.updatedCallback.emit(provider);
+  }
+
+  isGoogleIdP(idP: string): boolean {
```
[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -86,7 +101,26 @@ export class FormComponent implements OnInit {
        * Details: This calls the next method with the previous step, expecting
        * the next step to be returned, or a success or failure.
        ********************************************************************* */
-      const nextStep = await FRAuth.next(step, { tree: this.tree });
+      const selectIdCallback = step?.callbacks.find(
```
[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/app/app.module.ts
```diff
@@ -87,6 +90,9 @@ import { ChoiceComponent } from './features/journey/choice/choice.component';
     TodoIconComponent,
     ActionIconComponent,
     ChoiceComponent,
+    IdentityProviderComponent,
+    GoogleIconComponent,
+    AppleIconComponent,
```
cerebrl, line 95: Are we missing Facebook? It's one of the three we officially support, so we should probably have an example of it.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,35 @@ export class FormComponent implements OnInit {
...
+      if (selectIdCallback) {
+        const nameCallBacksInputs = step?.payload.callbacks.find(
+          (cb) => cb.type === 'NameCallback',
+        )?.input;
+        const idToken2Input = nameCallBacksInputs?.find((input) => input.name === 'IDToken2').value;
+
+        if (this.isIdentityProviderLogin(selectIdCallback) && idToken2Input !== '') {
+          selectIdCallback.payload.input[0].value = 'localAuthentication';
+        }
+      }
```
cerebrl, line 110: I'd really like to see some code comments in here helping the reader step through the logic of what we are trying to achieve here. It took me a bit to realize that this was using certain callback values to determine if we need to set `localAuthentication` as the value for the Identity Provider if they have filled in the name input.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -161,7 +199,17 @@ export class FormComponent implements OnInit {
...
+    const callBackInputValue = selectIdCallback?.payload.input[0].value as string;
+    if (selectIdCallback && callBackInputValue === '') {
+      selectIdCallback.payload.input[0].value = 'localAuthentication';
+    }
+    const redirectCallback = step.callbacks.find(
+      (callback) => callback.getType() === CallbackType.RedirectCallback,
+    );
+    if (redirectCallback) {
+      FRAuth.redirect(step);
+    }
```
cerebrl, line 212: This code is a bit dense. Can we get some code comments in here expressing intent? It's worth remembering the purpose of these sample apps, and that is to educate our customers on how to implement features. To do that, we need to a liberal amount of comments to guide the reader along.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -161,7 +199,17 @@ export class FormComponent implements OnInit {
    */
   handleStep(step?: FRStep) {
     this.step = step;
-
+    const selectIdCallback = step.callbacks.find((cb) => cb.payload.type === 'SelectIdPCallback');
```
cerebrl, line 202: Typo in the name, can we change this to `selectIdPCallback`?

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -161,7 +199,17 @@ export class FormComponent implements OnInit {
...
+    const callBackInputValue = selectIdCallback?.payload.input[0].value as string;
```
cerebrl, line 203: Since "callback" is one word, can we keep the case of the word consistent throughout our code. Right now, I see `callback` and `callBack`. I'd prefer the first: `callback`, and if it's the second word in the name: `Callback`, like `myCallback`.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,35 @@ export class FormComponent implements OnInit {
     this.submittingForm = true;
 
     try {
+      const selectIdCallback = step?.callbacks.find(
```
cerebrl, line 97: Same issue here, I'd like for this to be `selectIdPCallback`, so it properly reflects the actual name of the callback.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -161,7 +199,17 @@ export class FormComponent implements OnInit {
    */
   handleStep(step?: FRStep) {
     this.step = step;
-
+    const selectIdCallback = step.callbacks.find((cb) => cb.payload.type === 'SelectIdPCallback');
+    const callBackInputValue = selectIdCallback?.payload.input[0].value as string;
+    if (selectIdCallback && callBackInputValue === '') {
+      selectIdCallback.payload.input[0].value = 'localAuthentication';
+    }
```
cerebrl, line 206: Are we defaulting to `localAuthentication` here? If we are defaulting, do we also need the code above to essentially do the same thing? Or, if the above exists, does the code here need to exist? Maybe I'm misunderstanding the intention here, which is why I'd like to see comments, but it seems a bit redundant.

samples/angular-todo/src/app/features/journey/identityProvider/identity-provider.component.html
```diff
@@ -0,0 +1,14 @@
...
+    [ngClass]="{
+      'google-login': getDisplayName(provider) === idpNames.Google,
+      'apple-login': getDisplayName(provider) === idpNames.Apple,
+      'facebook-login': getDisplayName(provider) === idpNames.Facebook
+    }"
+    (click)="onSetProvider(provider.provider)"
+  >
+    <app-google-icon *ngIf="getDisplayName(provider) === idpNames.Google"></app-google-icon>
+    <app-apple-icon *ngIf="getDisplayName(provider) === idpNames.Apple"></app-apple-icon>
+    Sign in with {{ getDisplayName(provider) }}
```
cerebrl, line 12: Does Facebook default to this generic string template here, without an icon?

samples/angular-todo/src/callback.html
cerebrl, line 1: 👍

samples/angular-todo/src/environments/environment.ts
cerebrl, line 5: I'd really like to keep these values generic and/or use defaults found in a basic tenant. Could we switch these values back to generic or defaults?

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,35 @@ export class FormComponent implements OnInit {
     this.submittingForm = true;
 
     try {
+      const selectIdCallback = step?.callbacks.find(
+        (cb) => cb.payload.type === 'SelectIdPCallback',
+      ) as SelectIdPCallback;
+
+      if (selectIdCallback) {
+        const nameCallBacksInputs = step?.payload.callbacks.find(
```
ryanbas21, line 102: same as above here

samples/angular-todo/src/app/app.module.ts
juanManuel05: hi justin, facebook it is not here becasue if oyu have a look at the facebook button there's no icon to be added. It has no an "apple" or a "G" as apple and google does.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,35 @@ export class FormComponent implements OnInit {
...
+          (cb) => cb.type === 'NameCallback',
```
ryanbas21, line 103: same as above here. also, instead of `.input` we have a `getInputValue` method you can chain onto the end of this.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,35 @@ export class FormComponent implements OnInit {
...
+        )?.input;
+        const idToken2Input = nameCallBacksInputs?.find((input) => input.name === 'IDToken2').value;
```
ryanbas21, line 105: same as above here

samples/angular-todo/src/app/features/journey/form/form.component.ts
ryanbas21, line 203: Lets try to use the SDK api here with `getInputValue`

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -161,7 +199,17 @@ export class FormComponent implements OnInit {
    */
   handleStep(step?: FRStep) {
     this.step = step;
-
+    const selectIdCallback = step.callbacks.find((cb) => cb.payload.type === 'SelectIdPCallback');
+    const callBackInputValue = selectIdCallback?.payload.input[0].value as string;
+    if (selectIdCallback && callBackInputValue === '') {
+      selectIdCallback.payload.input[0].value = 'localAuthentication';
```
ryanbas21, line 205: same here

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -161,7 +199,17 @@ export class FormComponent implements OnInit {
...
+    }
+    const redirectCallback = step.callbacks.find(
```
ryanbas21, line 207: same here, we can getByType

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: of course not a problem. This has to be added because when pressing enter it selects the first identity provider by default so it was either disable the default behavior for submitting or keep the option up to submit the form when pressing enter.

juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -103,7 +139,9 @@ export class FormComponent implements OnInit {
           this.handleSuccess(nextStep);
           break;
         case 'Step':
-          this.handleStep(nextStep);
+          {
+            this.handleStep(nextStep);
+          }
```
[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: i had to add this workaround because it crashes if it was a local authentication, but now is no longer needed. It is taken care of at line 100. removed it, made a few tests and works

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: removed

samples/angular-todo/src/app/features/journey/identityProvider/identity-provider.component.html
juanManuel05: it has no icon before text
Screenshot 2023-08-08 at 15 18 26

samples/angular-todo/src/environments/environment.ts
juanManuel05: It could set it back, however it'd be a different journey with no option for IdPs.
thoughts??

cerebrl: Yeah, let's set it back. We should always have the sample apps point to the default trees: `Login` and `Registration`. As we build out new features, and support new nodes/callbacks, we will continue to default the sample apps to these trees. We are just adding these features in order to support an expanded set of nodes _if_ the customer uses them when experimenting with the sample apps.

samples/angular-todo/src/app/app.module.ts
cerebrl: Okay, if that's consistent with Facebook's display rules, then I'm good with it.

samples/angular-todo/src/app/features/journey/identityProvider/identity-provider.component.html
cerebrl: Thanks for the screenshot. Can you remove the default borders around the buttons, so they follow the same style convention as the other buttons in the app? Only the Google button needs a border, since the background is white, so let's use the same style is the "outline button" that should already be defined in the Bootstrap library. The outline button should have the same style as the other form elements.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,35 @@ export class FormComponent implements OnInit {
     this.submittingForm = true;
 
     try {
+      const selectIdCallback = step?.callbacks.find(
+        (cb) => cb.payload.type === 'SelectIdPCallback',
```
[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/app.module.ts
ryanbas21: https://developers.facebook.com/docs/facebook-login/userexperience#buttondesign

it seems like you can use either, maybe we should be consistent with the other buttons?

samples/angular-todo/src/app/features/journey/form/form.component.ts
ryanbas21, line 97: should we not use the getCallbacksOfType here also?

samples/angular-todo/src/app/shared/idps.enums.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Copyright (c) 2020 ForgeRock. All rights reserved.
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+/**
+ * Types of identity providers
+ */
+enum IdpNames {
```
ryanbas21, line 14: I have a preference that we use Unions instead of enums, as a best practice. Can we convert this to a union type?

ryanbas21 (changes requested):

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,37 @@ export class FormComponent implements OnInit {
     this.submittingForm = true;
 
     try {
+      /** This was to be added because when logging in locally by pressing enter it selects the first identity provider by default
```
ryanbas21, line 97: I think we should format the comments in the same style as below.

For example this comment may look like

```
/** *********************************************************************
       * SDK INTEGRATION POINT
       * Summary: Handle selection of IdpCallbacks
       * ----------------------------------------------------------------------
       * Details: If we do not initially select the IdpCallback, when pressing enter,
       * local authentication will automatically be submitted. 
       * This allows us to submit the form with the proper selection of an 
       * identity provider rather than local authentication.                      
 /** *********************************************************************

```
       ********************************************************************* */

juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -79,14 +94,37 @@ export class FormComponent implements OnInit {
...
+       *
+       */
+      const selectIdPCallback = step?.getCallbacksOfType(CallbackType.SelectIdPCallback);
+
+      if (selectIdPCallback?.length > 0) {
+        const nameCallBacksInputs = step?.getCallbackOfType(CallbackType.NameCallback);
+        const idToken2Input = nameCallBacksInputs?.getInputValue('IDToken2');
+
+        if (this.isIdentityProviderLogin(selectIdPCallback[0]) && idToken2Input !== '') {
+          selectIdPCallback[0].payload.input[0].value = 'localAuthentication';
```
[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.html
```diff
@@ -154,6 +154,27 @@ <h1 class="text-center fs-2 mb-3" *ngIf="title">{{ title }}</h1>
...
               ></app-kba>
 
+              <!--
+                Handle a KbaCreateCallback using the Kba component.
+                Emitted questions and answers are used to set the values on the callback.
+              -->
+              <app-identity-provider
+                *ngSwitchCase="'SelectIdPCallback'"
+                [callback]="$any(callback)"
+                [name]="callback?.payload?.input?.[0]?.name"
```
ryanbas21, line 164: getInput?

samples/angular-todo/src/app/features/journey/form/form.component.ts
ryanbas21, line 110: getInputValue?

samples/angular-todo/src/app/features/journey/form/form.component.html
juanManuel05: it's the same for the other components. Followed the standard

ryanbas21 (approved):
Thanks for hanging in there through the review.

cerebrl (approved):

PR #381, opened 2023-08-03
feat(javascript-sdk): add-logger-functionality

ryanbas21:
# JIRA Ticket
https://bugster.forgerock.org/jira/browse/SDKS-2125)
# Description
Implemented a logger function which is a singleton. It contains some private variables that a user can dynamically change. A change to the singleton would last for the lifetime of it, so if a user wants to change the logger back to defaults, they would need to either kill the singleton, or reinitialize the defaults by passing console through the api.

I gave log levels a "rating" which is internal to the logger module. These allow for us to accept custom log types later on - we would adjust the type slightly to accept an argument and a log level rating.

This made assigning the hierarchy simpler and not need multiple checks. 

# Type of Change

# How Has This Been Tested?

Unit tested / integration tested.

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


packages/javascript-sdk/src/logger/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    addLogger({
+      logFn = console.log,
+      warnFn = console.warn,
+      errorFn = console.error,
+    }: LoggerFunctions) {
+      frLog = logFn;
+      frWarn = warnFn;
+      frError = errorFn;
+    },
+    shouldLog() {
```
cerebrl, line 24: Not a huge fan of multi-word properties, if we can avoid them. Would `enabled` work as the method name? What's more, I'm not normally a fan of using "[getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)" and "setters", but I feel it might be fairly useful here. That way, we can also avoid making the dev call the function. They could just `frLogger.enabled` and get back the boolean?

packages/javascript-sdk/src/logger/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    },
+    log(...msgs: unknown[]) {
+      if (this.shouldLog()) {
+        frLog(...msgs);
+      }
+    },
+  };
+}
+
+const FRLogger = customLogger();
```
cerebrl, line 45: Since this is no longer a class, should we start with a capital letter? Does it matter? Is there a difference?

packages/javascript-sdk/src/logger/index.ts
```diff
@@ -0,0 +1,47 @@
+import Config from '../config/index';
+
+interface LoggerFunctions {
+  warnFn?: (...msgs: unknown[]) => void;
+  errorFn?: (...msgs: unknown[]) => void;
+  logFn?: (...msgs: unknown[]) => void;
+}
+
+function customLogger() {
```
cerebrl, line 9: So, this is a pretty new convention we're setting here. In the case of the `TokenStore`, we passed the custom store object in the `Config.set` method. Since this is new, can we do a quick cost-benefit of adding this new convention?

cerebrl (commented):
Not sure if you knew about this, but here's what the other platforms have written on this topic: https://wikis.forgerock.org/confluence/display/SDKS/SDK+Logging. That's the parent page, but there's two child pages as well in the left nav. Also, here's an epic that was done by Jey a while ago, for reference: https://bugster.forgerock.org/jira/browse/SDKS-1823.

packages/javascript-sdk/src/config/interfaces.ts
```diff
@@ -12,6 +12,8 @@ import type { ActionTypes } from './enums';
 import type { FRCallbackFactory } from '../fr-auth/callbacks/factory';
 import type { Tokens } from '../shared/interfaces';
 
+type LogLevel = 'none' | 'info';
```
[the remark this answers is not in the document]
ryanbas21: So we only have `warn` calls in the sdk, i wasn't sure what the thought was on "enabling" and "disabling" and how much control we really want to enable. So if we have a defined type of logging we want to do, we can add types and functionality.

packages/javascript-sdk/src/logger/index.ts
ryanbas21: We can move this into the closure if thats your suggestion, thats easy enough

packages/javascript-sdk/src/logger/index.ts
ryanbas21: So this was my thought process.

I started with an abstract class, because it made sense that we didn't need to instantiate any class and it reduced the footprint of changes in the SDK.

Then when we suggested adding a custom logger, an abstract class didn't really work unless we modify the config level code. Once we pass in a custom logger, theres no real way of passing arguments through an abstract config because its static, and it doesn't get invoked. 

I didn't really know if we wanted to expand Config beyond the general "ForgeRock" AM configuration settings and I felt logging wasn't AM Configuration but more preference of the consumer.

The downside being there needs to be an explicit call of the FRLogger in this case to configure logging, but its dynamic and can be changed on the fly without reinstantiating several other redundant variables.

If we keep it in the config, we have to reinitialize the config with all the required arguments, like `serverConfig` because without it, this code breaks here:

```
 private static isValid(options: ConfigOptions): boolean {
    return !!(options && options.serverConfig);
  }
```

We can even remove loglevel from Config if we wanted and just keep it in the new instantiated singleton, i'm open to it either way. I just thought theres more of a chance a consumer will never call FRLogger and they are more likely to call Config.set, set the logging they want (or not) and never look at it again.

The more I wrote code the more I felt we wanted to keep these variables private so that warn / log / and error could be changed but they were never really leaked out of the implementation, so a closure was the only true way to do that.


I think theres more flexibility to an object approach, but we could i suppose, instantiate the class the same way, the token store doesn't accept a custom store so i don't see how similar it is in this sense, if it did the Token Store I don't think would be able to be abstract?

packages/javascript-sdk/src/logger/index.ts
ryanbas21: I'm not sure if you want them to be able to use the logger to verify if logging is enabled?

Right now if someone wants to check on Logging, they can do so through `Config.logLevel` but we could move all of this to the Logger and even remove `LogLevel` from `Config`, is that your thought?

packages/javascript-sdk/src/logger/index.ts
ryanbas21: Probably not good convention so i can change this to lower case.

packages/javascript-sdk/src/config/interfaces.ts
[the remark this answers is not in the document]
cerebrl: I think we would like to align with the other SDKs, plus, we have a few rare cases of `info` and `error`, so let's use the level to represent the logging with more inclusivity. By that, I mean:

- `info`: will log everything, including `info`, `warn` and `error`
- `warn`: will log `warn` and `error`
- `error`: will log `error` only
- `none`: will not log

I hope that make sense. You might want to confirm this with Andy. Our desire is to slowly start aligning logging and error handling at least at a fundamental level with each SDK.

packages/javascript-sdk/src/logger/index.ts
cerebrl: Chatted on Slack. Let's see how the changes look.

packages/javascript-sdk/src/config/interfaces.ts
```diff
@@ -12,12 +12,25 @@ import type { ActionTypes } from './enums';
...
   type: ActionTypes;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   payload: any;
 }
-
+/**
+ * Custom Logger for logger
+ */
+interface LoggerFunctions<
+  W = (...msgs: unknown[]) => void,
```
ryanbas21, line 26: defaulting this to a console.log/warn/error type, but this will allow a client to pass in their loggers type if needed, it may not be much different but i'm not sure how loggers type themselves and maybe this will provide more ergonomics if a client needs to use an FRLogger on their own?

cerebrl (changes requested):
A few issues, but overall, it's coming together:

- Looks like you're missing the log statements on `http-client/helpers.ts`
- The new convention is a bit worrisome on the `Config` class
- I think `info` is missing in the logger and config

Let's make sure to get all the instances of `console` used in the SDK, so we don't have any leaks.

cerebrl (approved):

PR #384, opened 2023-08-22
feat(angular-todo): webAuthn support for Angular

juanManuel05:
# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2643

# Description
Add WebAuthn support for Angular todo sample app

# Type of Change

# How Has This Been Tested?

- Click on sign in button
- If the journey handles username and/or password, they will requested. As a final step, a biometric input will be requested.
- if succesfully logged, it redirects to the home page.
- If failure, it redirects to the login page.


samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
```diff
@@ -0,0 +1,117 @@
...
+import { Component, Input, OnInit } from '@angular/core';
+import {
+  FRAuth,
+  FRLoginFailure,
+  FRLoginSuccess,
+  FRStep,
+  TokenManager,
+  UserManager,
+} from '@forgerock/javascript-sdk';
+import FRWebAuthn from 'packages/javascript-sdk/src/fr-webauthn';
```
ryanbas21, line 20: can we use the above import?

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
```diff
@@ -0,0 +1,117 @@
...
+import { WebAuthnStepType } from 'packages/javascript-sdk/src/fr-webauthn/enums';
```
ryanbas21, line 21: same as above

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -73,6 +87,7 @@ export class FormComponent implements OnInit {
   identityProviders: IdPValue[];
   code: string;
   state: string;
+  isShowSignIn = true;
```
ryanbas21, line 90: Can we call this variable `showSignIn`?

ryanbas21 (changes requested):

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
juanManuel05: @ryanbas21 got this message "Projects cannot be imported by a relative or absolute path, and must begin with a npm scope"

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
juanManuel05: same message as avobe

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -203,6 +218,28 @@ export class FormComponent implements OnInit {
...
+    const textOutputCallbacks = step.getCallbacksOfType(CallbackType.TextOutputCallback);
+
+    if (textOutputCallbacks.length > 0) {
+      this.isShowSignIn = false;
+    } else {
+      this.isShowSignIn = true;
+    }
+
+    /**
+     * Since webAuthn flow must be handling as a completely different feature, it is necessary to check if it is a webAuthn type step,
```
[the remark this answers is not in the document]
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
ryanbas21: thats exactly what the above import is doing

juanManuel05: done

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
```diff
@@ -0,0 +1,117 @@
...
+          this.header = 'Registering your device';
+          this.step = await FRWebAuthn.register(this.step);
+          break;
+
+        case WebAuthnStepType.Authentication:
+          this.isWebAuthnRegistration = false;
+          this.message = 'Use your device to verify your identity';
+          this.header = 'Verifying your identity';
+          this.step = await FRWebAuthn.authenticate(this.step);
+          break;
```
cerebrl, line 58: We should probably have the SDK INTEGRATION POINT JS Doc comment for this block of code, yeah?

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -203,6 +218,28 @@ export class FormComponent implements OnInit {
    */
   handleStep(step?: FRStep) {
     this.step = step;
+
+    /**
+     * If the step contains TextOutputCallback, it means it has webAuthn callbacks so the sign-in button must be hidden as
+     * webAuthn will provide both biometric registration and biometric authentication log-in option buttons.
```
cerebrl, line 224: This comment is technically correct, and I don't want our customers to be misled by it. If a step has a `TextOutputCallback`, it doesn't mean it's WebAuthn. Our WebAuthn node does use `TextOutputCallback`, but it could also use `MetadataCallback`. Because of this, the presence of this callback doesn't indicate anything. The only way we can know this is a WebAuthn step is by using the dedicated `isWebAuthnType` method.

samples/angular-todo/src/app/features/journey/form/form.component.ts
```diff
@@ -203,6 +218,28 @@ export class FormComponent implements OnInit {
...
+     */
+    const textOutputCallbacks = step.getCallbacksOfType(CallbackType.TextOutputCallback);
+
+    if (textOutputCallbacks.length > 0) {
+      this.isShowSignIn = false;
+    } else {
+      this.isShowSignIn = true;
+    }
```
cerebrl, line 232: What does `isShowSignIn` represent?

samples/angular-todo/src/app/views/login/login.component.html
```diff
@@ -11,18 +11,25 @@
 <div class="cstm_container_v-centered container-fluid d-flex align-items-center">
   <div class="w-100">
     <app-back-home></app-back-home>
+    <ng-container *ngIf="!isWebAuthn; else webAuthn">
```
cerebrl, line 14: A few things here:

1) Because I'm not super familiar with Angular, this isn't immediately obvious what's going on to me, but the model is hardcoding `this.isWebAuthn` to `false`, so the condition here of `!isWebAuthn` will always evaluate to `true` on the first pass. Yeah?

Then, you bind a function called `onSetIsWebAuthn` to an event named `isWebAuthnn` (is the double "n" necessary?). When this event is emitted, you're flipping the `this.isWebAuthn` to true? This makes the WebAuthn component to render. Is that correct?

2) Either way, I don't believe we should the logic for WebAuthn step at this level. The login view is a bit too specific to make these decisions. We should be making this decision a bit deeper in the component tree, like in `form.component.ts`. I say this because WebAuthn can be used in both login AND registration. As it is right now, WebAuthn couldn't be used for any registration flow.

What's more, I think this logic would prevent WebAuthn from being the first node in a tree, which that too is possible. Let me know if you have any questions about this.

samples/angular-todo/src/app/features/journey/form/form.component.ts
cerebrl: Oh, I think I figured it out. Can we rename this button to be something more specific and accurate to its purpose? Something like `showSubmitButton`?

cerebrl (changes requested):
This PR looks like it's coming together well, but we still have a bit of work to do. Let me know if you have any questions about my comments. Thanks, Juan.

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
juanManuel05: Regarding the first quesiton, it's done i'd misunderstood the question. Sorry, dont get the last comment

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.html
```diff
@@ -0,0 +1,38 @@
...
+Copyright (c) 2021 ForgeRock. All rights reserved.
+This software may be modified and distributed under the terms
+of the MIT license. See the LICENSE file for details.
+-->
+
+<ng-container
+  *ngTemplateOutlet="
+    success ? successMessage : failure ? failureMessage : loading;
+    context: { message: { message: message, header: header } }
+  "
```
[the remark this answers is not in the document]
juanManuel05: oh yes, that's basically a similar logic being applied at the form component.  It works as an IF statement, depending on the value of those variables 'success' and 'failure' it will load a template (which is below), either successMessage or failureMessage or loading template. At the same time, by passing the 'context' atribute, it sends an object to be used by the template.

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: yes, no problem. Can you talk it to ryan too? he'd asked me to name showSignIn(i made the change already locally)

samples/angular-todo/src/app/views/login/login.component.html
juanManuel05: 1. the first assumpiton is correct. Depending on the emitted value login component will be loading or not the webAuthn component. 
2. it'd take some changes as i made all of the implementation based on the premise that it's either the webAuthn funcionality or the form component as this comment says  "Since WebAuthn is being handled as its own unique component, it is necessary to check if we have WebAuthn type step. if so an output event is emitted to the log-in component to load the webAuhtn component."
Before startibng implementing, after a talk i had to ryan we decided to go this way, and of course i accepted as a valid option. I'd like to discuss it the three of us if possible.

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: got it, shall i remove it then?

cerebrl: Ugh, sorry. I have a typo in my original comment. I meant to write, "This comment is NOT technically correct ... " Also, I'm not talking directly about the comment itself, but also the logic we are using below. Since we have access to `step` and already pull in `FRWebAuthn`, I'd rather use use the `FRWebAuthn.getWebAuthnStepType` method to derive states of UI, rather than the presence of `TextOutputCallback`. The presence of `TextOutputCallback` is not a good indicator of much, unfortunately.

samples/angular-todo/src/app/features/journey/form/form.component.ts
cerebrl: @ryanbas21: since Juan is using `isShowSignIn` as a hint to the UI regarding the display of a submit button, I'd like it to be more descriptive of intent. The sign in flow is not the only use case where the submit button may not be displayed, so naming it `showSignIn` is a bit misleading. Plus, a casual dev may not deduce what this does or doesn't do by reading this code alone.

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
cerebrl: Sorry, let me restate my comment for better clarity. Being that we are try to use JSDoc style comments through this code base to help provide hints to the developer using this to learn how to use the SDK, I think it would be good to have a comment here.

The comment could explain what may not be immediately obvious, like the try catch, the fact that the step is mutated, even when an error is thrown. Let's try to help the developer with comments to ensure they don't get stuck on the same thing you got tripped up on. Make sense?

These comments look like this:

```js
/** *********************************************************************
 * SDK INTEGRATION POINT
 * Summary: Short description
 * ----------------------------------------------------------------------
 * Details: Longer explanation
 ************************************************************************* */
```

Does that make sense? We try to liberally use the comments throughout the sample app codebase.

samples/angular-todo/src/app/views/login/login.component.html
cerebrl: Hey Juan, sorry for the confusion. Please go ahead and push your changes, so we can see your solution. Remember, as long as the last commit in your Git history is **_your_** commit, you can use `git commit --amend` and include your changes in that existing commit. That way, you don't have to add additional commits when addressing PR comments. 

The only thing is when you amend a commit (or rebase), it rewrites history, so when you push, you have to force push. This is fine as long as you're only rewriting your own commit, and you're force pushing to a branch that is not shared with anyone (like if you're pair programming with someone and sharing a branch). To force push, you'd use `git push -f origin webAuthn`. 

If, for any reason, you'd like to know why force pushing is necessary, or what rewriting history means, please read this article I wrote. Others find it quite helpful: https://cerebralideas.com/blog/rewriting-git-history

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
juanManuel05: ah got it,  not a problem i will add a comment explaning what the code does.

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/views/login/login.component.html
juanManuel05: thanks i will ready it! commiting my changes soon

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/webAuthn/webAuthn.component.ts
juanManuel05: done

samples/angular-todo/src/app/features/journey/form/form.component.ts
juanManuel05: done

samples/angular-todo/src/app/views/login/login.component.html
juanManuel05: done

samples/angular-todo/src/app/views/login/login.component.html
```diff
@@ -11,18 +11,25 @@
 <div class="cstm_container_v-centered container-fluid d-flex align-items-center">
   <div class="w-100">
     <app-back-home></app-back-home>
+    <ng-container>
+      <div class="card shadow-sm p-5 mb-2 w-100">
+        <div class="cstm_form-icon align-self-center mb-3">
+          <ng-container *ngIf="!isWebAuthn; else fingerPrint">
```
cerebrl, line 17: Do we need to make the same change to `register.component.html` and `register.component.ts`? Usually, the WebAuthn would be optionally registered in the registration flow. So, both the login flow and the registration flow should handle WebAuthn, which is why I discouraged the core logic from being in this view. This logic looks like it just handles icon changes, which is great, but it would be nice if this same treatment was on the register flow.

samples/angular-todo/src/app/features/journey/confirmation/confirmation.component.html
```diff
@@ -0,0 +1,16 @@
...
+
+Copyright (c) 2021 ForgeRock. All rights reserved.
+This software may be modified and distributed under the terms
+of the MIT license. See the LICENSE file for details.
+-->
+<div class="cstm_form d-grid gap-2">
+  <ng-container *ngFor="let option of callback.getOptions()">
+    <button class="cstm_form btn btn-primary w-100" (click)="onSetOptionValue(option)">
+      {{ option }}
+    </button>
```
cerebrl, line 14: Let's turn this into a select element or radio inputs like I mention in my Gist here: https://gist.github.com/cerebrl/766f7df333bcf6f180fa0ac8de4010d2. That way, we don't have to worry about figuring out if we need to render the submit button or not in this case.

samples/angular-todo/src/app/views/login/login.component.html
juanManuel05: this is only intended to switch the key-icon to the finger-print icon? is it necessary for the registration? we dont just fill out a form when registering? may be wrong. It culd be easily added tho, just asking.
image
==========================
image

juanManuel05: this is what i meant
image

cerebrl (approved):
LGTM, thank you, Juan, for your hard work here :)

ryanbas21 (approved):
Look great!

PR #386, opened 2023-08-29
fix(javascript-sdk): expose-logger-functions-type

ryanbas21:
# Description

expose the LoggerFunctions type because a client could use that in their code to define the custom logger

# Type of Change


packages/javascript-sdk/src/index.ts
```diff
@@ -124,6 +125,7 @@ export {
   HttpClient,
   KbaCreateCallback,
   LocalStorage,
+  LoggerFunctions,
```
cerebrl, line 128: Since this is just a type, can we export it in the above block that's dedicated to exporting just types?

cerebrl (approved):

PR #394, opened 2023-10-11
feat(reactjs-todo): add WebAuthn support for react todo sample app

juanManuel05:
# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2771

# Description
[React JS] Add WebAuthn support for react todo sample app


samples/reactjs-todo/client/components/journey/confirmation.js
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <div className="cstm_form-floating mb-3">
+      <select
+        onChange={setOptionValue}
+        id={inputName}
+        className={`cstm_form-select form-select bg-transparent ${state.theme.textClass} ${state.theme.borderClass}`}
+      >
+        {options.map(function (option, idx) {
+          return (
+            <option key={idx} value={option}>
```
ryanbas21, line 50: this isn't a super important note, but in terms of best practice, we probably want to use something other than `idx` for the key. this is pretty minor but maybe we can use something on `option` instead, that is more unique.

Theres reasoning from a react perspective behind it, but its pretty negligible in my opinion and not super important.

samples/reactjs-todo/client/components/journey/form.js
```diff
@@ -121,6 +125,10 @@ export default function Form({ action, bottomMessage, followUp, topMessage }) {
         return <TermsConditions callback={cb} inputName={name} key={name} />;
       case 'KbaCreateCallback':
         return <Kba callback={cb} inputName={name} key={name} />;
+      case 'TextOutputCallback':
+        return <TextOutput key={`textOutput-${idx}`} />; //For textOutput call backs, 'input' field comes empty which leads to a  unique-key-prop error
```
ryanbas21, line 129: `// For TextOutput callbacks, 'input' field comes empty which leads to a unique-key-prop error`

samples/reactjs-todo/client/components/journey/text-output.js
```diff
@@ -0,0 +1,23 @@
...
+ * @returns {Object} - React component object
+ */
+
+/** 
+ component added to render purposes, at this point this callback renders no element
+*/
+
+// eslint-disable-next-line no-unused-vars
+export default function TextOutput({ inputName }) {
+  return null;
```
ryanbas21, line 22: sorry if i missed context on this, but why are we doing this?

samples/reactjs-todo/client/components/journey/webAuthn.js
```diff
@@ -0,0 +1,63 @@
...
+ * @function WebAuthn - Footer React component
+ * @returns {Object} - React component object
+ */
+export default function WebAuthn({ step, setSubmissionStep }) {
+  const webAuthnStep = FRWebAuthn.getWebAuthnStepType(step);
+  const [state, setState] = useState({
+    message: '',
+    header: '',
+  });
+  useEffect(() => {
```
ryanbas21, line 24: I see we are using a functional setState in this effect. I'm curious why we made this distinction?

It doesn't seem like its possible to set the state multiple times before the render in this control flow? I think we should make it a regular setState unless i'm missing something to make it seem a bit less complex.

Functional setState comes with a little more complexity.

samples/reactjs-todo/client/components/journey/text-output.js
juanManuel05: some of the callbacks that webauthn steps have are of the type TextOutput and i need to handle them even if it's just rendering null, otherwise it crashes rendering the next error. I'm following the same approach i did for the angular sample app, just handling but not rendering anything.
image
image

samples/reactjs-todo/client/components/journey/webAuthn.js
juanManuel05: you mean two different states like this,one for the header and another for message?. Going thru react docs i saw i could "combine" two values into one unique state and went for it. Is it a bad practice? I could use an explanation for upcoming features
image

samples/reactjs-todo/client/components/journey/text-output.js
ryanbas21: Ok, thank you, what is in the textoutput callback in this case?

samples/reactjs-todo/client/components/journey/webAuthn.js
ryanbas21: I am referring to this

   ```
         setState((state) => ({
            ...state,
            message: 'Your device will be used to verify your identity',
          }));
   ```
   
   why not
   
   `setState({...state,message: 'Your device will be used to verify your identity'})`

samples/reactjs-todo/client/components/journey/text-output.js
juanManuel05: image

samples/reactjs-todo/client/components/journey/webAuthn.js
juanManuel05: i just tried and it enters into a loop. Reacts aks for the state to be added as a dependency and when i test it, it enters into a loop, trying to constantly log in. 
I assumed i was doing this to have the value for the previous state, like a reducer in redux pattern,
```
setState((state) => ({
         ...state,
         message: 'Your device will be used to verify your identity',
}))
```;

File: samples/reactjs-todo/client/components/utilities/loading.js [hunk 5]
```diff
@@ -18,15 +18,15 @@ import { AppContext } from '../../global-state';
...
+export default function Loading({ classes, message, header }) {
   const [state] = useContext(AppContext);
-
   return (
     <div className="container">
       <p className={classes}>
         <span className="d-flex justify-content-center my-2">
           <span className="cstm_loading-spinner spinner-border text-primary" role="status"></span>
         </span>
+        <span className="d-flex justify-content-center fw-bolder">{header}</span>
```
ryanbas21, line 29: This needs a default and something that doesnt render a header when it isn't passed in. we use this component without header prop in other areas

juanManuel05: done

samples/reactjs-todo/client/components/journey/text-output.js
cerebrl: I don't think we should `return null;` here. We should print what is present in the callbacks. Is there a reason we aren't just rendering the string that is contained in the text output callback?

samples/reactjs-todo/client/views/login.js
```diff
@@ -30,7 +31,7 @@ export default function Login() {
         <BackHome />
         <Card>
           <div className="cstm_form-icon  align-self-center mb-3">
-            <KeyIcon size="72px" />
+            {!state.isWebAuthn ? <KeyIcon size="72px" /> : <FingerPrintIcon size="72px" />}
```
cerebrl, line 34: If I'm reading the code well, I think this is the change that necessitates the need for the global state. But, I think a better solution is to move this icon to a component that has the appropriate context for displaying the right one without the need for leaking WebAuthn state into the global space.

We could easily move this icon to the `Form` component, and keep this component clean. Or, even better still, we could move the icon to the `WebAuthn` component.

samples/reactjs-todo/client/views/register.js
```diff
@@ -35,7 +36,7 @@ export default function Register() {
         <BackHome />
         <Card>
           <div className="cstm_form-icon align-self-center mb-3">
-            <NewUserIcon size="72px" />
+            {!state.isWebAuthn ? <NewUserIcon size="72px" /> : <FingerPrintIcon size="72px" />}
```
cerebrl, line 39: Same argument goes for this as well.

samples/reactjs-todo/client/components/journey/webAuthn.js
```diff
@@ -0,0 +1,54 @@
...
+ */
+import { FRWebAuthn, WebAuthnStepType } from '@forgerock/javascript-sdk';
+import React, { useEffect, useState } from 'react';
+import Loading from '../utilities/loading';
+
+/**
+ * @function WebAuthn - Footer React component
+ * @returns {Object} - React component object
+ */
+export default function WebAuthn({ step, setSubmissionStep }) {
```
cerebrl, line 18: I'd advocate for making this more into a complex view, broadening its responsibility. This is what makes this a named stage. Let's move all things WebAuthn into here. Since WebAuthn is deterministic and not dynamic, we don't have to concern ourselves with it being composed with other callbacks or unknown contexts.

cerebrl (changes requested):
This is looking like it's coming together quite well. I just have a few changes that I'd like to recommend to keep the code clean and maintainable. I think the architectural suggestions I'm making here are why I didn't fully complete the feature on our React test app.

samples/reactjs-todo/client/components/journey/text-output.js
juanManuel05: in this case, i'd print "login" or "0". Does it make sense?. we did the same fot the angular sample app

samples/reactjs-todo/client/components/utilities/loading.js
```diff
@@ -18,15 +18,15 @@ import { AppContext } from '../../global-state';
...
+export default function Loading({ classes, message, header }) {
   const [state] = useContext(AppContext);
-
   return (
     <div className="container">
       <p className={classes}>
         <span className="d-flex justify-content-center my-2">
           <span className="cstm_loading-spinner spinner-border text-primary" role="status"></span>
         </span>
+        {header ? <span className="d-flex justify-content-center fw-bolder">{header}</span> : null}
```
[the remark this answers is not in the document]
juanManuel05: it's here as it's part of the loading messge(attached an image) so it's necesary it to be here. Once the webAuthn comes into action the loading component is gone.

samples/reactjs-todo/client/global-state.js
```diff
@@ -119,11 +120,13 @@ export function useGlobalStateMgmt({ email, isAuthenticated, prefersDarkTheme, u
       email: mail,
       theme,
       username: name,
+      isWebAuthn: isWebAuthn,
```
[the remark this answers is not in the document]
juanManuel05: ok i will remove it. It is used in login and register to render either key-icon or finger-print-icon
image

samples/reactjs-todo/client/components/utilities/loading.js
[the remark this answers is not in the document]
juanManuel05: moved into web-authn

samples/reactjs-todo/client/components/journey/webAuthn.js
```diff
@@ -0,0 +1,54 @@
...
+          payload: {
+            message: 'user cancelled webAuthn',
+          },
+        });
+      }
+    }
+    performWebAuthn();
+  }, [step, webAuthnStep, setSubmissionStep]);
+
+  return <Loading message={state.message} header={state.header} />;
```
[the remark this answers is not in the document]
juanManuel05: done

samples/reactjs-todo/client/views/register.js
juanManuel05: same reply as above

samples/reactjs-todo/client/components/journey/webAuthn.js
juanManuel05: Overall, i addresed all of the comments and moved the logic into here. There's just one thing i don't yet know how to work out, and this is how to let the parent know it's web-authn flow and hide the key-icon?(above comment)

samples/reactjs-todo/client/views/login.js
cerebrl: There's no reason why icons in general have to be rendered in this top level view. If that's what's causing the struggle, let's just move them where we have the context?

Why don't we just remove them entirely from this view and put them into the `Form` component. Then the `Form` component can make all the decisions as it has the context of what should and shouldn't render without need to spread journey state into the global state. Does that make sense?

Also, the `Form` component has the information of whether the user is logging in or registering because we pass that `action.type` into the `Form` component. So, it already has the needed information to make a decision on the key versus user icon as well.

samples/reactjs-todo/client/components/journey/webAuthn.js
cerebrl: Let me know if the earlier comment clears up how to achieve this goal. Thanks!

samples/reactjs-todo/client/views/login.js
juanManuel05: got it! i will work on this workaround, let you know any block.

samples/reactjs-todo/client/components/journey/webAuthn.js
juanManuel05: done

samples/reactjs-todo/client/components/journey/web-authn.js
```diff
@@ -0,0 +1,66 @@
...
+        } else {
+          setState({
+            header: 'Verifying your identity',
+            message: 'Use your device to verify your identity',
+          });
+          await FRWebAuthn.authenticate(step);
+        }
+        setSubmissionStep(step);
+      } catch (e) {
+        setSubmissionStep({
```
ryanbas21, line 42: If we set this as the submission step, should we not have something handling this type of error in the `journey` hook? 

What is the expected behavior when someone cancels a webauthn step? A webauthn node has a `unsupported, `success`, `failure` and `client error`  possible case. In the case of cancel, is this `failure`?

if we fail should we be redirecting to a default journey? Is this what happens when we throw this error? I feel like we're missing the handling of this error and just throwing it to throw rather than handling it. I don't believe we have a standard error like this that we handle by default

cerebrl: You are correct, Ryan. We should submit the step as is — e.g. (`setSubmissionStep(step)`) — just like in the success case. The `step` gets modified in the WebAuthn method before it throws. This gives the dev an opportunity to handle the error, or just submit it as is.

The client itself never declares `LoginFailure`, that's the responsibility of the server. So, we shouldn't ever directly use this `LoginFailure` step type.

samples/reactjs-todo/client/components/journey/text-output.js
cerebrl: Oh, I see what you mean. Go ahead and use the `TextOutputCallback` as it's intended. Rather than returning `null`, which just hides the message, let's render the text. 

Now, that will result in two headers essentially, "Sign In" and "Login", so let's fix that by designing a better Message Node. The message itself should ask a user a question, not act as a title, and then the options should represent the user's answer. 

Does that make sense?

BTW, if you can, avoid "resolving a conversation" when it still contains an open question or request. Thanks.

cerebrl (approved):

PR #397, opened 2023-10-23
update readme to point to localhost for todos API

mattkey-forgerock:
Updated Readme for the React-todos sample, to use non-ssl localhost for the resource API instead of https://api.example.com:9443. Also updated .env sample to be more in-line with sample provided in Backstage docs.


cerebrl (commented):
Thanks, Matt! This looks good to me, but I'll leave this to @ryanbas21 for final approval.

PR #399, opened 2023-10-26
feat(reactjs-todo): add social login for react todo sample app

juanManuel05:
# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2762

# Description
add social login for react todo sample app
# Type of Change

### DIFF
--- a/samples/angular-todo/src/app/features/journey/form/form.component.html
+++ b/samples/angular-todo/src/app/features/journey/form/form.component.html
@@ -196,7 +196,7 @@ of the MIT license. See the LICENSE file for details.
               </app-confirmation>
 
               <!--
-                Handle a RedirectCallback using the Kba component.
+                Handle a RedirectCallback using the loading component.
                 Emitted questions and answers are used to set the values on the callback.
               -->
               <app-loading
--- a/samples/reactjs-todo/client/components/icons/apple-icon.js
+++ b/samples/reactjs-todo/client/components/icons/apple-icon.js
@@ -1,0 +1,40 @@
+/*
+ * forgerock-sample-web-react
+ *
+ * apple-icon.js
+ *
+ * Copyright (c) 2021 ForgeRock. All rights reserved.
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+import React from 'react';
+
+/**
+ * @function AppleIcon - React component for the user icon representing the apple icon
+ * @param {Object} props - React props object
+ * @param {string} props.classes - A string of classnames to be set on component
+ * @param {string} props.size - A string representing the intended size of the rendering
+ * @returns {Object} - React JSX Object
+ */
+export default function AppleIcon({ classes = '', size = '24px' }) {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      className={classes}
+      height={size}
+      width={size}
+      viewBox="0 0 256 256"
+    >
+      <g fill="#070000" strokeMiterlimit="10" style={{ mixBlendMode: 'normal' }}>
+        <path d="M0 256V0h256v256z"></path>
+      </g>
+      <path
+        fill="#fff"
+        strokeMiterlimit="10"
+        d="M32.5 44c-1.778 0-3.001-.577-4.08-1.086C27.38 42.424 26.481 42 25 42c-1.481 0-2.38.424-3.42.914C20.501 43.423 19.278 44 17.5 44 13.174 44 6 34.071 6 23.5 6 16.49 10.832 11 17 11c2.027 0 3.259.581 4.346 1.093C22.378 12.58 23.27 13 25 13s2.622-.42 3.654-.907C29.741 11.581 30.973 11 33 11c2.664 0 5.033.982 7.042 2.921a1.5 1.5 0 01-.288 2.376C37.438 17.644 36 20.499 36 23.75c0 3.661 2.004 6.809 4.986 7.831.391.134.709.423.879.799.171.375.18.805.023 1.188C39.461 39.515 35.424 44 32.5 44zm-7-34a1.499 1.499 0 01-1.494-1.637c.012-.123.303-3.045 2.593-5.382C28.753.781 30.85.127 31.081.059a1.5 1.5 0 011.904 1.652c-.034.241-.389 2.436-2.232 4.899-1.973 2.636-4.791 3.322-4.91 3.35-.114.027-.229.04-.343.04z"
+        transform="scale(5.33333)"
+      ></path>
+    </svg>
+  );
+}
--- a/samples/reactjs-todo/client/components/journey/form.js
+++ b/samples/reactjs-todo/client/components/journey/form.js
@@ -9,7 +9,7 @@
  */
 
 import React, { Fragment, useEffect, useContext, useReducer } from 'react';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
 
 import Boolean from './boolean';
 import { DEBUGGER } from '../../constants';
@@ -27,11 +27,12 @@ import Unknown from './unknown';
 import Button from './button';
 import TextOutput from './text-output';
 import Confirmation from './confirmation';
-import { FRWebAuthn, WebAuthnStepType } from '@forgerock/javascript-sdk';
+import { FRWebAuthn, WebAuthnStepType, FRAuth, CallbackType } from '@forgerock/javascript-sdk';
 import WebAuthn from './web-authn';
 import KeyIcon from '../../components/icons/key-icon';
 import NewUserIcon from '../../components/icons/new-user-icon';
 import FingerPrintIcon from '../../components/icons/finger-print-icon';
+import IdentityProvider from './identity-provider';
 
 /**
  * @function Form - React component for managing the user authentication journey
@@ -56,6 +57,16 @@ export default function Form({ action, bottomMessage, followUp, topMessage }) {
   const [form] = useReducer(treeReducer, treeReducer(null, action));
   // Used for redirection after success
   const navigate = useNavigate();
+  const [params] = useSearchParams();
+
+  // Get the code and state from the URL query parameters
+  const codeParam = params.get('code');
+  const stateParam = params.get('state');
+
+  let resumeUrl = '';
+  if (codeParam && stateParam) {
+    resumeUrl = window.location.href;
+  }
 
   /**
    * Custom "hook" for handling form orchestration
@@ -63,7 +74,7 @@ export default function Form({ action, bottomMessage, followUp, topMessage }) {
   const [
     { formFailureMessage, renderStep, submittingForm, user },
     { setSubmissionStep, setSubmittingForm },
-  ] = useJourneyHandler({ action, form });
+  ] = useJourneyHandler({ action, form, resumeUrl });
 
   /**
    * If the user successfully authenticates, let React complete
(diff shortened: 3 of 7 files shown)


samples/reactjs-todo/client/components/journey/form.js
```diff
@@ -172,7 +191,32 @@ export default function Form({ action, bottomMessage, followUp, topMessage }) {
...
+    const selectIdPCallback = renderStep?.getCallbacksOfType(CallbackType.SelectIdPCallback);
+
+    if (selectIdPCallback?.length > 0) {
+      const nameCallBacksInputs = renderStep?.getCallbackOfType(CallbackType.NameCallback);
+      const idToken2Input = nameCallBacksInputs?.getInputValue('IDToken2');
+
+      if (isIdentityProviderLogin(selectIdPCallback[0]) && idToken2Input !== '') {
+        selectIdPCallback[0].setInputValue('localAuthentication', 0);
+      }
+    }
```
cerebrl, line 219: I'm not sure I understand the intent here. Can you elaborate on what this is doing? 

Just speaking generically, there are a few things that give me hesitation here:

1. It seems like we have logic at the form level that should be at the callback level. We should be able to achieve a cleaner implementation, similar to what we have in the Login Framework. Please checkout the implementation in that project to get a sense of how we do it there.
2. We have logic that uses values that may be inconsistent between journey designs. The reliance of `IDToken2` will not always be reliable as it's an implementation detail of the journey itself, rather than a guaranteed interface.

samples/reactjs-todo/client/components/journey/form.js
```diff
@@ -132,12 +143,20 @@ export default function Form({ action, bottomMessage, followUp, topMessage }) {
...
     }
   }
-
+  function isIdentityProviderLogin(callback) {
+    return callback
+      ?.getOutputByName('providers', [])
+      .filter((provider) => provider.provider !== 'localAuthentication').length > 0
+      ? true
+      : false;
+  }
```
cerebrl, line 159: I'm a bit confused here. Is the intention to identify if this is social login by testing if it has 1 or more social identity providers, once you filter out `localAuthentication`? I ask because if there's a `SelectIdPCallback` in the step, this will always be true. The `SelectIdPCallback` will never have just `localAuthentication` in it. So, the existence of `SelectIdPCallback` is itself enough to detect social login.

samples/reactjs-todo/client/components/icons/apple-icon.js
```diff
@@ -0,0 +1,52 @@
...
+      viewBox="0 0 256 256"
+    >
+      <g
+        fill="#070000"
+        strokeMiterlimit="10"
+        fontFamily="none"
+        fontSize="none"
+        fontWeight="none"
+        textAnchor="none"
+        style={{ mixBlendMode: 'normal' }}
```
cerebrl, line 36: I believe all these properties can be removed, yeah?

samples/reactjs-todo/client/components/icons/apple-icon.js
```diff
@@ -0,0 +1,52 @@
...
+      </g>
+      <path
+        fill="#fff"
+        strokeMiterlimit="10"
+        d="M32.5 44c-1.778 0-3.001-.577-4.08-1.086C27.38 42.424 26.481 42 25 42c-1.481 0-2.38.424-3.42.914C20.501 43.423 19.278 44 17.5 44 13.174 44 6 34.071 6 23.5 6 16.49 10.832 11 17 11c2.027 0 3.259.581 4.346 1.093C22.378 12.58 23.27 13 25 13s2.622-.42 3.654-.907C29.741 11.581 30.973 11 33 11c2.664 0 5.033.982 7.042 2.921a1.5 1.5 0 01-.288 2.376C37.438 17.644 36 20.499 36 23.75c0 3.661 2.004 6.809 4.986 7.831.391.134.709.423.879.799.171.375.18.805.023 1.188C39.461 39.515 35.424 44 32.5 44zm-7-34a1.499 1.499 0 01-1.494-1.637c.012-.123.303-3.045 2.593-5.382C28.753.781 30.85.127 31.081.059a1.5 1.5 0 011.904 1.652c-.034.241-.389 2.436-2.232 4.899-1.973 2.636-4.791 3.322-4.91 3.35-.114.027-.229.04-.343.04z"
+        fontFamily="none"
+        fontSize="none"
+        fontWeight="none"
+        textAnchor="none"
+        transform="scale(5.33333)"
```
cerebrl, line 48: These too should be removable, yeah?

samples/reactjs-todo/client/components/journey/identity-provider.js
```diff
@@ -0,0 +1,67 @@
...
+  function getDisplayName(idp) {
+    switch (idp.uiConfig.buttonDisplayName) {
+      case 'Google':
+        return 'Google';
+      case 'Facebook':
+        return 'Facebook';
+      case 'Apple':
+        return 'Apple';
+      default:
+        return 'Google';
```
cerebrl, line 44: We should probably default to a generic text only button. Falling back to Google feels a bit odd here.

samples/reactjs-todo/client/components/journey/identity-provider.js
```diff
@@ -0,0 +1,67 @@
...
+        <button
+          className={providersClass[provider.uiConfig.buttonDisplayName]}
+          key={idx}
+          onClick={() => onSetProvider(provider.provider)}
+        >
+          {providerName === 'Google' ? (
+            <GoogleIcon size="24px" />
+          ) : providerName === 'Apple' ? (
+            <AppleIcon size="24px" />
+          ) : null}{' '}
```
cerebrl, line 61: Is there a reason we have Facebook in the logic above, but not implemented here?

cerebrl (changes requested):

samples/reactjs-todo/client/components/journey/identity-provider.js
juanManuel05: we've had the same question for the angular implementation. I did it this way because the idp service didn't come along with a facebook icon. Just followed the design

samples/reactjs-todo/client/components/icons/apple-icon.js
juanManuel05: removing it breaks the svg 
image

samples/reactjs-todo/client/components/journey/form.js
juanManuel05: right, removed!

samples/reactjs-todo/client/components/journey/form.js
juanManuel05: thanks for the comment, let's see if i can be clear enough. When logging locally via user/password, the step coming back from AM comes along with 3 types of CB (selectIDPcallback, nameCallback and passwordCallback). I'd say due to the implementation of the journey/AM, the selectIdPCallback .input[0] for the payload has the value of the first idp, in this case facebook, so it leads to a failure authentication as if it is local log-in that value "selectIdPCallback --> input[0]  -->**idToken1**" must be "localAuthentication". So having said this, i proceed to do this exactly:
1) i check the existence of a selectIdpCallback, since this "misbehavior" only occurs for the selectIdpCallback
2) if so, grab for the nameCallback to detect if it's local auth. If this value is empty(the user didn't enter a name for that input in the form), it is social-login log-in, don't do anything.
3) Check for the the value of its input, if it has a value we want it to be 'localAuthentication' as we're logging in bia user/password **through a selectIdpCallback** , hence set the value manually. Keep in mind that it could be a selectIdpCallback and still be a user/password log-in, which is why i only set the value manually if nameCallback has a value. I could've logged in via google and nameCallback would be emtpy, as the user didn't enter the username, in that case i wouldn't do anything
I attach an image where i'm logging in via username/password, the stepRender has a value of "facebook-web" for the selectIdPCallback .input[0], as mentioned before it is supposed to be 'localAuthentiction', then i proceed to set it.
image

juanManuel05: UPDATE: we could go for this workaround. I took parte of the logic from the login widget but it's still necessary to check for the NameCallBack input. In the image, the nameCallbackInput is empty, which means it is accesing via idp(google-web-juan) and we don't need to set the provider to localAuthentication, we leave it as it is
image

**solution proposed:**
image

samples/reactjs-todo/client/components/icons/apple-icon.js
cerebrl: Which of the above properties are required? I wouldn't think al the `font*` related properties are required, yeah? We can keep the ones required, but trim the ones that aren't.

samples/reactjs-todo/client/components/icons/apple-icon.js
cerebrl: Here too. Maybe the `transform` property is needed, but I wouldn't think the `font*` related properties would be needed.

samples/reactjs-todo/client/components/journey/form.js
cerebrl: > When logging locally via user/password, the step coming back from AM comes along with 3 types of CB (selectIDPcallback, nameCallback and passwordCallback). I'd say due to the implementation of the journey/AM, the selectIdPCallback .input[0] for the payload has the value of the first idp, in this case facebook, so it leads to a failure authentication as if it is local log-in that value "selectIdPCallback --> input[0] -->idToken1" must be "localAuthentication".

Sure, I believe the most optimal way of handling this is what we do in the `select-idp.svelte` component here:https://github.com/ForgeRock/forgerock-web-login-framework/blob/beta/src/lib/journey/callbacks/select-idp/select-idp.svelte#L50. 

The logic is simply this:

1. If the `SelectIdPCallback` has `localAuthentication` option, immediately change the default value to local option
2. If the user fills out the username and password without interaction of the providers, then no need to change from local option
3. If the user wants to use Google or whatever, then they will click on the provider altering the default we set
4. If the user clicks on on a social provider, they can't "change their mind" as the redirect will take them away from the Widget, and they will have to return, resetting the default back to local

Does that make sense? This way, callbacks don't need to know about each other and the stage level doesn't have to have knowledge about specific callbacks.

samples/reactjs-todo/client/components/icons/apple-icon.js
juanManuel05: done

samples/reactjs-todo/client/components/icons/apple-icon.js
juanManuel05: done

samples/reactjs-todo/client/components/journey/identity-provider.js
juanManuel05: done

samples/reactjs-todo/client/components/journey/form.js
juanManuel05: i took the logic from this very component and if i only used that "if" the social provider log-in breaks. In the example below i logged in via google. i do have localAuthentication (1), therefore it will enter into the "if", and the provider(in thise case google-web-juan) (2) will be overwritten to "localAuthentication" but we dont want that, we want it to remain as "google-web-juan" as we're accesing via google. (4)
image

image

cerebrl: This is likely because you have this logic inside the "bare" component, rather than within a `useEffect` with `[]` dependencies. For every change in state, the components will rerun themselves, and you'll keep setting the provider back to `localAuthentication` over and over. What you want is to just set that value at first render, but NOT every subsequent render, allowing the user to select and keep the desired option.

Can you move this logic for setting the default `localAuthentication` provider setting into the `identity-provider` component, and wrap it in a `useEffect`, so it only runs once on initial render? I have a feeling that will fix your issue.

@ryanbas21 Does my above logic make sense to you. If it does, can you help out here?

ryanbas21 (approved):

samples/angular-todo/src/app/features/journey/form/form.component.html
```diff
@@ -196,7 +196,7 @@ <h1 class="text-center fs-2 mb-3" *ngIf="title">{{ title }}</h1>
               </app-confirmation>
 
               <!--
-                Handle a RedirectCallback using the Kba component.
```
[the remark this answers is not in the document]
juanManuel05: oh yes, it should've been like this in the first place. When i did the angular implementation i copied the text from another component and forgot to rename it.

samples/reactjs-todo/client/components/journey/identity-provider.js
```diff
@@ -0,0 +1,67 @@
...
+  const identityProviders = callback
+    .getProviders()
+    .filter((provider) => provider.provider !== 'localAuthentication');
+
+  function onSetProvider(provider) {
+    if (DEBUGGER) debugger;
+    callback.setProvider(provider);
+  }
+
+  function getDisplayName(idp) {
```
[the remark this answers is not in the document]
juanManuel05: agree, done

cerebrl (approved):

PR #400, opened 2023-11-01
feat(javascript-sdk): include raw response to OAuth token return

cerebrl:
## DO NOT MERGE

# JIRA Ticket

[SDKS-2790](https://bugster.forgerock.org/jira/browse/SDKS-2790)

# Description

Adjust `OAuth2Client` class to attach the raw response from the server as an additional property that propagates through to the application layer and also is stored along with the traditional tokens.

## Details

- add raw response to returned value in OAuth2Client
- this raw response is the original serialized JSON body
- this new property will be persisted with the tokens in WebStorage

# Type of Change

# How Has This Been Tested?

Manually tested with the e2e autoscript apps (screenshot included) as well as added a test assertion for our automated e2e test suite as well.

Screenshot of manual test with localStorage with `rawResponse` containing the original response body

CleanShot 2023-11-01 at 15 38 23

# Definition of Done

Check all that apply

- [x] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [x] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread

### DIFF
--- a/e2e/autoscript-apps/src/authn-oauth/autoscript.ts
+++ b/e2e/autoscript-apps/src/authn-oauth/autoscript.ts
@@ -88,6 +88,7 @@ function autoscript() {
         } else {
           throw new Error('Session_Error');
         }
+        console.log(tokens.rawResponse);
         return tokens;
       }),
       rxDelay(delay),
--- a/e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
+++ b/e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
@@ -15,14 +15,28 @@ test.describe('Test OAuth login flow', () => {
   test(`should login successfully and then log out with`, async ({ page, browserName }) => {
     const { messageArray, networkArray } = await setupAndGo(page, browserName, 'authn-oauth/');
 
+    let rawResponse = '';
+
     // Test assertions
     // Test log messages
-    expect(messageArray.includes('OAuth login successful')).toBe(true);
-    expect(messageArray.includes('Logout successful')).toBe(true);
-    expect(messageArray.includes('Calling authorize endpoint')).toBe(true);
-    expect(messageArray.includes('Calling access token exchange endpoint')).toBe(true);
-    expect(messageArray.includes('Get user info from OAuth endpoint')).toBe(true);
-    expect(messageArray.includes('New OAuth tokens retrieved')).toBe(true);
+    expect(messageArray.includes('OAuth login successful'), 'oauth success').toBe(true);
+    expect(messageArray.includes('Logout successful'), 'logout success').toBe(true);
+    expect(messageArray.includes('Calling authorize endpoint'), 'call /authorize').toBe(true);
+    expect(
+      messageArray.includes('Calling access token exchange endpoint'),
+      'call /access_token',
+    ).toBe(true);
+    expect(messageArray.includes('Get user info from OAuth endpoint'), 'call /userinfo').toBe(true);
+    expect(messageArray.includes('New OAuth tokens retrieved'), 'tokens received').toBe(true);
+
+    // Test rawResponse on token object
+    messageArray.forEach((message) => {
+      if (message.includes('access_token')) {
+        rawResponse = message;
+      }
+    });
+
+    expect(rawResponse.includes('access_token')).toBe(true);
 
     // Test network requests
     // Make sure revoke request is made twice, one for force renew and one for logout
--- a/e2e/autoscript-suites/src/suites/authz-txn-basic-json.test.ts
+++ b/e2e/autoscript-suites/src/suites/authz-txn-basic-json.test.ts
@@ -16,9 +16,21 @@ test.describe('Test Transaction Authorization flow using JSON response', () => {
     const { messageArray } = await setupAndGo(page, browserName, 'authz-txn-basic-json/');
 
     // Test assertions
-    expect(messageArray.includes('IG resource requires additional authorization')).toBe(true);
-    expect(messageArray.includes('Request to IG resource successfully responded')).toBe(true);
-    expect(messageArray.includes('Starting authentication with composite advice')).toBe(true);
-    expect(messageArray.includes('Continuing authentication with composite advice')).toBe(true);
+    expect(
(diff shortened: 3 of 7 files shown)


e2e/autoscript-apps/src/authn-oauth/autoscript.ts
```diff
@@ -88,6 +88,7 @@ function autoscript() {
         } else {
           throw new Error('Session_Error');
         }
+        console.log(tokens.rawResponse);
```
cerebrl, line 91: Adding a log of the raw response, so the test suite can pick it up and ensure the original response has the correct values.

e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
```diff
@@ -15,14 +15,28 @@ test.describe('Test OAuth login flow', () => {
...
+    ).toBe(true);
+    expect(messageArray.includes('Get user info from OAuth endpoint'), 'call /userinfo').toBe(true);
+    expect(messageArray.includes('New OAuth tokens retrieved'), 'tokens received').toBe(true);
+
+    // Test rawResponse on token object
+    messageArray.forEach((message) => {
+      if (message.includes('access_token')) {
+        rawResponse = message;
+      }
+    });
```
cerebrl, line 37: This is the new code that tests the existence of the raw response, which uses snake_case. Because this is a live capable test, I just want to test the existence of keys, not values.

e2e/autoscript-suites/src/suites/authn-oauth.lc.test.ts
```diff
@@ -15,14 +15,28 @@ test.describe('Test OAuth login flow', () => {
...
-    expect(messageArray.includes('New OAuth tokens retrieved')).toBe(true);
+    expect(messageArray.includes('OAuth login successful'), 'oauth success').toBe(true);
+    expect(messageArray.includes('Logout successful'), 'logout success').toBe(true);
+    expect(messageArray.includes('Calling authorize endpoint'), 'call /authorize').toBe(true);
+    expect(
+      messageArray.includes('Calling access token exchange endpoint'),
+      'call /access_token',
+    ).toBe(true);
+    expect(messageArray.includes('Get user info from OAuth endpoint'), 'call /userinfo').toBe(true);
+    expect(messageArray.includes('New OAuth tokens retrieved'), 'tokens received').toBe(true);
```
cerebrl, line 30: I'm adding secondary messages here, so if the test fails, it will print the message to help inform which expectation failed. Without this, the test just says, "expected true and got false" or something like that.

e2e/autoscript-suites/src/suites/authz-txn-basic-json.test.ts
```diff
@@ -16,9 +16,21 @@ test.describe('Test Transaction Authorization flow using JSON response', () => {
...
+      'successful response',
+    ).toBe(true);
+    expect(
+      messageArray.includes('Starting authentication with composite advice'),
+      'start auth with advice',
+    ).toBe(true);
+    expect(
+      messageArray.includes('Continuing authentication with composite advice'),
+      'continue with advice',
+    ).toBe(true);
```
cerebrl, line 34: This randomly failed, and I was motivated to add the secondary messages here to, but there's no actual code change here.

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -160,6 +160,7 @@ abstract class OAuth2Client {
     };
 
     const response = await this.request('accessToken', undefined, false, init, options);
+    const responseClone = response.clone();
```
cerebrl, line 163: Cloning the response here since we want to both grab the response text as well as create parsed JSON out of it.

e2e/autoscript-apps/src/authn-oauth/autoscript.ts
ryanbas21: Should probably use the Logger class?

packages/javascript-sdk/src/oauth2-client/index.ts
ryanbas21: Can you point me to what you mean? I don't understand why the `clone` is necessary?

packages/javascript-sdk/src/oauth2-client/index.ts
```diff
@@ -185,6 +186,7 @@ abstract class OAuth2Client {
       idToken: responseObject.id_token,
       refreshToken: responseObject.refresh_token,
       tokenExpiry: tokenExpiry,
+      rawResponse: await responseClone.text(),
```
ryanbas21, line 189: so the basic ask is to just include the entire response object in addition to how we structure it normally.

i'm curious if we should include this as an _opt in_ configuration? it may be overkill to add as a config but i also think we are adding a lot to what was previously a very thin api that could add noise for many who just don't need this info

ryanbas21 (changes requested):
thoughts for discussion

packages/javascript-sdk/src/oauth2-client/index.ts
cerebrl: Good question! This is something that took me a bit to wrap my head around (I had to do this a lot in the Token Vault plugin). The tl;dr of is any time you call a method off of `body`, you "attach" a "reader" to the underlying `ReadableStream` of the `body`. Only one reader can be attached to a stream at a time, so if you have two different places in the code that need to "attach" to the stream of the `body`, then you have to `clone` it.

The other location where we read the `body` is here: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/oauth2-client/index.ts#L301.

Ref for the above: https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams#attaching_a_reader

ryanbas21: wow havent used this before at all. interesting stuff

PR #466, opened 2024-10-02
fix: update-signals-sdk

ryanbas21:
# JIRA Ticket

Please link jira ticket here

# Description

# Type of Change

# How Has This Been Tested?

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


packages/ping-protect/package.json
```diff
@@ -20,6 +20,6 @@
   },
   "types": "./dist/index.ts.d.ts",
   "dependencies": {
-    "@forgerock/javascript-sdk": "workspace:^"
+    "@forgerock/javascript-sdk": "^4.8.2"
```
ryanbas21: why this change?

packages/ping-protect/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@forgerock/ping-protect",
-  "version": "4.6.0",
+  "version": "4.6.0-beta.1",
```
ryanbas21: ?

cerebrl (approved):

packages/javascript-sdk/src/fr-auth/callbacks/ping-protect-initialize-callback.ts
```diff
@@ -27,15 +27,27 @@ class PingOneProtectInitializeCallback extends FRCallback {
    */
   public getConfig() {
     const config = {
+      // Required parameter
       envId: this.getOutputByName<string>('envId', ''),
+
+      // Optional parameters
+      agentIdentification: this.getOutputByName<boolean>('agentIdentification', false),
+      agentTimeout: this.getOutputByName<number>('agentTimeout', 0),
```
adamheath-fr: When testing these changes the fact that these values are being defaulted to `0` is causing issues as this is setting the port and timeout to `0` (Meaning the agent does not get connected to).

When the values are not set in the node the expectation is that we should be using the default values, which is in line with the signals SDK init call expecting these values to be present **ONLY** if the default values are not desired - see documentation here https://apidocs.pingidentity.com/pingone/native-sdks/v1/api/#pingone-protect-sdk-for-web:~:text=//%20If%20you%20have,9400)%0A%20%20%20%20%20%20%20%20//%20agentPort%3A%208800 

If default values have to be provided in this class, perhaps we should use the Signals SDK default values here instead?

Default values for these configuration options are:
- `agentPort` = 9400
- `agentTimeout` = 1000

(Default values inferred from documentation here: https://docs.pingidentity.com/pingone/strong_authentication_mfa/p1_using_the_workforce_trust_agent.html#:~:text=the%20agent%20timeout%20(-,1000%20ms,-)%2C%20use%20the%20PingID )

cerebrl: Okay, I can make the changes. Rather than dealing with default values, I'll just remove the config property if it's not on the callback. Would that work better?

adamheath-fr: That would be great - keeps things in line with the Signals expected SDK behaviour and if they update their defaults we don't need any additional update.

Thanks

adamheath-fr:
Tested these changes with the new beta and things are looking good - thanks for the swift turnaround on this. ✅

adamheath-fr (approved):

PR #492, opened 2024-11-18
test(automation): add more unit, integration and e2e test automation

cerebrl:
# JIRA Ticket

[Jira ticket SDKS-3386](https://pingidentity.atlassian.net/browse/SDKS-3386)

# Description

Add automation for unit, integration and e2e tests.

# Type of Change

- [x] Add tests and automation

# How Has This Been Tested?

Automation

# Definition of Done

Check all that apply

- [x] Coded to standards.
- [ ] Code peer-reviewed.
- [x] Unit tests are written.
- [x] Integration tests are written.
- [x] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread


e2e/davinci-app/main.ts
```diff
@@ -11,13 +11,11 @@ import flowLinkComponent from './components/flow-link.js';
 import socialLoginButtonComponent from './components/social-login-button.js';
 
 const config = {
-  // clientId: '7544ef78-5c01-48c4-a9a5-6338c10c7f87',
-  clientId: '724ec718-c41c-4d51-98b0-84a583f450f9', // Luis' tenant
+  clientId: '724ec718-c41c-4d51-98b0-84a583f450f9',
```
ryanbas21, line 14: Do we want these to be PR'd or should we begin a migration to secrets / .env file?

Honestly - if we don't need these to be secrets I think that's the best way to go about it.

Without a more scaleable solution for us as we scale tenants and packages and what not, its going to be really hard to maintain various configuration settings (unless we have some secret solution that we currently do not have).

e2e/davinci-app/main.ts
```diff
@@ -168,7 +171,7 @@ const config = {
    * It returns an unsubscribe function that you can call to stop listening
    */
   davinciClient.subscribe(() => {
-    const node = davinciClient.getClientInfo();
+    const node = davinciClient.getClient();
```
ryanbas21, line 174: This could just be me, but when i read this for some reason my brain thought this would be an async function.

packages/davinci-client/src/types.ts
```diff
@@ -1,3 +1,5 @@
+import 'immer'; // Side-effect needed only for getting types in workspace
```
ryanbas21, line 1: This feels like a smell. We would have to mark this as a `side effect` in the package.json.

If this is purely a type related need there should be better ways of doing it.

A) I think all types should be default included unless a `types` field is included. So i think if we have a `types` field in the package.json maybe we should remove it, or we should add this type to the field.

Is the expected behavior just that the `immer` types are scoped in this module?

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -34,14 +35,15 @@ export type ActionCollectorTypes =
...
+  error: string | null;
   type: ActionCollectorTypes;
   id: string;
   name: string;
   output: {
     key: string;
     label: string;
     type: string;
-    url?: string;
+    url?: string | null;
```
ryanbas21, line 46: This is a bit of a redundant type?

e2e/davinci-app/main.ts
cerebrl: This isn't a secret, just a client ID in our test tenant. Though, I want to switch these tests to use the mock API we have, so this will probably go away.

e2e/davinci-app/main.ts
cerebrl: Yeah, originally, it was just `clientInfo` or something, but then more methods kept getting added to the API, and it started getting a bit crowded with just having nouns, so I added the verb prefix. These `get*` method don't result in network requests, just "getting" a portion of the existing, stored state.

packages/davinci-client/src/lib/collector.types.ts
cerebrl: Anytime I'm explicitly using setting a value to be "no value", I'm using `null`. This helps differentiate between `undefined`, which a interpret as a value was expected, but not value was provided, and `null` which means we intentionally provided a "no value" value. I hope that makes sense. But, I think you have some changes that may render this particular change moot, yeah?

ryanbas21: Yeah, I may have changed this. The redundant part is the `?` in the url.

packages/davinci-client/src/types.ts
ryanbas21: I guess i'm confused as what the purpose is, Why are we trying to populate the module scope with the immer types?

cerebrl: I remember having a conversation with you about this a while ago. I think it's from instances of the `createSlice`. Without getting the `immer` types into the scope, the variable assigned a `createSlice` has to have an explicit, generic (not `<T>` type generic, but just generic as in plain) type on it, which then removes the implicit type, which contains the specific attributes of that particular slice. I hope that makes sense.

Maybe you'll understand it better than me, so I'd encourage you to pull this yourself and remove the immer stuff. Checkout the slices to see the error that should be present in your IDE.

ryanbas21 (approved):

PR #493, opened 2024-11-18
fix(davinci-client): types-and-wellknown

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/jira/software/c/projects/SDKS/boards/1234?selectedIssue=SDKS-3594

# Description
use wellknown endpoint to get endpoints available. Refactor of types

# Type of Change

# How Has This Been Tested?

# Definition of Done

Check all that apply

- [ ] Coded to standards.
- [ ] Code peer-reviewed.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.
- [ ] Demo to PO and team.

## Documentation

- [ ] Acceptance criteria met
- [ ] Spell-check run
- [ ] Peer reviewed
- [ ] Proofread

### DIFF
--- a/packages/davinci-client/src/lib/client.store.ts
+++ b/packages/davinci-client/src/lib/client.store.ts
@@ -29,16 +29,15 @@ export async function davinci({ config }: { config: DaVinciConfig }) {
     throw Error('wellknown endpoint required as part of the `config.serverOptions` ');
   }
 
-  const { data: endpoints } = await store.dispatch(
+  const { data: openIdResponse } = await store.dispatch(
     wellknownApi.endpoints.wellknown.initiate(config.serverConfig.wellknown),
   );
 
-  if (!endpoints) {
+  if (!openIdResponse) {
     throw Error('error fetching wellknown response');
   }
 
-  store.dispatch(configSlice.actions.setEndpoints(endpoints));
-  store.dispatch(configSlice.actions.set(config));
+  store.dispatch(configSlice.actions.set({ ...config, endpoints: openIdResponse }));
 
   return {
     // Pass store methods to the client
--- a/packages/davinci-client/src/lib/client.store.ts
+++ b/packages/davinci-client/src/lib/client.store.ts
@@ -11,7 +11,8 @@ import { configSlice } from './config.slice.js';
  */
 import type { DaVinciConfig } from './config.types.js';
 import type { DaVinciAction, DaVinciRequest } from './davinci.types.js';
-import type { SingleValueCollector } from './collector.types.js';
+import type { SingleValueCollectors } from './collector.types.js';
+import { wellknownApi } from './wellknown.api.js';
 
 /**
  * Create a client function that returns a set of methods
@@ -24,6 +25,19 @@ import type { SingleValueCollector } from './collector.types.js';
 export async function davinci({ config }: { config: DaVinciConfig }) {
   const store = createClientStore();
 
+  if (!config.serverConfig.wellknown) {
+    throw Error('wellknown endpoint required as part of the `config.serverOptions` ');
+  }
+
+  const { data: endpoints } = await store.dispatch(
+    wellknownApi.endpoints.wellknown.initiate(config.serverConfig.wellknown),
+  );
+
+  if (!endpoints) {
+    throw Error('error fetching wellknown response');
+  }
+
+  store.dispatch(configSlice.actions.setEndpoints(endpoints));
   store.dispatch(configSlice.actions.set(config));
 
   return {
@@ -68,7 +82,7 @@ export async function davinci({ config }: { config: DaVinciConfig }) {
      * @param {SingleValueCollector} collector - the collector to update
      * @returns {function} - an async function to call for updating collector value
      */
-    update: (collector: SingleValueCollector) => {
+    update: (collector: SingleValueCollectors) => {
       const { id } = collector;
       return function (value: string, index?: number) {
         store.dispatch(nodeSlice.actions.update({ id, value, index }));
--- a/packages/davinci-client/src/lib/collector.utils.ts
+++ b/packages/davinci-client/src/lib/collector.utils.ts
@@ -20,11 +20,11 @@ import type { DaVinciField } from './davinci.types';
  * @param {ActionCollectorTypes} [collectorType] - Optional type of the ActionCollector.
  * @returns {ActionCollector} The constructed ActionCollector object.
  */
-export function returnActionCollector(
+export function returnActionCollector<CollectorType extends ActionCollectorTypes>(
   field: DaVinciField,
   idx: number,
-  collectorType?: ActionCollectorTypes,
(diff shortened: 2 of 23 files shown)


packages/davinci-client/src/lib/config.slice.ts
```diff
@@ -40,22 +40,15 @@ export const configSlice = createSlice({
...
+        state.responseType = action.payload.responseType;
       } else {
-        state.serverConfig = {
-          baseUrl: action.payload.serverConfig?.baseUrl.concat('/'),
-        };
+        state.responseType = 'code';
       }
+      state.scope = action.payload.scope || 'openid';
+    },
+    setEndpoints(state, action: PayloadAction<Endpoints>) {
```
cerebrl, line 50: I'm not entirely sold on having this be a different reducer. We should be able to just include the endpoints in the regular `set` reducer, since it's all going in the same place and we already have a president with the JS SDK: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/config/interfaces.ts#L91 (although named `paths` and not `endpoints`).

Unless, I'm missing something that requires this be a separate reducer.

packages/davinci-client/src/lib/node.slice.ts
```diff
@@ -253,7 +255,9 @@ export const nodeSlice = createSlice({
       // TODO: Improve type
       const newState = state as Draft<NextNode>;
 
-      newState.client.collectors = nodeCollectorReducer(newState.client.collectors, action);
+      const a = nodeCollectorReducer(newState.client.collectors, action);
+
+      newState.client.collectors = a;
```
cerebrl, line 260: Did you mean to commit this?

packages/davinci-client/src/lib/wellknown.api.ts
```diff
@@ -0,0 +1,27 @@
...
+
+      transformResponse: (value: OpenIdResponse): Endpoints => {
+        return {
+          authorize: value.authorization_endpoint,
+          issuer: value.issuer,
+          introspection: value.introspection_endpoint,
+          tokens: value.token_endpoint,
+          userinfo: value.userinfo_endpoint,
+        };
+      },
```
cerebrl, line 24: If we are going to store the endpoints in the `config` slice, I think I'd prefer to keep the response unmodified here. That way it's cached untouched, and the `config` slice's reducer maps the original response properties appropriately. This is what we do in the DaVinci API.

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -20,11 +20,11 @@ import type { DaVinciField } from './davinci.types';
  * @param {ActionCollectorTypes} [collectorType] - Optional type of the ActionCollector.
  * @returns {ActionCollector} The constructed ActionCollector object.
  */
-export function returnActionCollector(
+export function returnActionCollector<CollectorType extends ActionCollectorTypes>(
   field: DaVinciField,
   idx: number,
-  collectorType?: ActionCollectorTypes,
-): ActionCollector {
+  collectorType: CollectorType,
```
cerebrl, line 26: Can you elaborate on how this improves our typing?

cerebrl (commented):
This looks good overall, just have some questions left in comments.

packages/davinci-client/src/lib/node.slice.ts
ryanbas21: Ah you're right was debugging. will back out that part.

packages/davinci-client/src/lib/wellknown.api.ts
ryanbas21: What do you mean by "its cache is untouched" The `transformResponse` occurs before caching is what I understand?

packages/davinci-client/src/lib/collector.utils.ts
ryanbas21: discussed on slack

packages/davinci-client/src/lib/config.slice.ts
ryanbas21: I can make it happen in `set` thats fine. I've left it as endpoints unless you feel strongly about `paths`. I've personally never loved "paths" naming but I'm happy to discuss something that may be better than `endpoints`.

cerebrl: That's fine. We can use `endpoints` here; I agree about `paths`.

packages/davinci-client/src/lib/wellknown.api.ts
cerebrl: If you transform the response as you are doing here, you are going to cache the transformation result, not the original response from the server. I'd personally prefer to cache the original.

CleanShot 2024-11-19 at 13 21 00

ryanbas21: So at what stage would you rather the transformation happened? 

- Would we pass the entire openId response into the reducer and have the reducer pull out the endpoints?
- Would we do it ourselves before calling `set` action? 

Its a minor thing i'm just unsure where the best place to do this is if not in `transformResponse`. Doesn't matter much to me either way

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -24,7 +25,19 @@ import type { SingleValueCollector } from './collector.types.js';
 export async function davinci({ config }: { config: DaVinciConfig }) {
   const store = createClientStore();
 
-  store.dispatch(configSlice.actions.set(config));
+  if (!config.serverConfig.wellknown) {
+    throw Error('wellknown endpoint required as part of the `config.serverOptions` ');
+  }
+
+  const { data: openIdResponse } = await store.dispatch(
```
ryanbas21, line 32: So this gets the openIdResponse from the wellknown call.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -24,7 +25,19 @@ import type { SingleValueCollector } from './collector.types.js';
...
+    wellknownApi.endpoints.wellknown.initiate(config.serverConfig.wellknown),
+  );
+
+  if (!openIdResponse) {
+    throw Error('error fetching wellknown response');
+  }
+
+  store.dispatch(configSlice.actions.set({ ...config, endpoints: openIdResponse }));
```
ryanbas21, line 40: Pass in the entire wellknown response here, open to changing the key `endpoints` to something else, @cerebrl like `openIdConfiguration`?

packages/davinci-client/src/lib/wellknown.api.ts
cerebrl: I'd say just pass the original response to the config slice. You don't have to transform or normalize the response from the wellknown endpoint since it follows a spec. All responses from the wellknown should be the same. That means you can just pass the original response to the config slice and have it map what it needs to the stored config.

PR #494, opened 2024-11-22
chore: update-nx

ryanbas21:
# Description
Update infrastructure - no ticket.


ryanbas21:
self-merging becase its baasic update & password update

PR #546, opened 2025-06-30
fix: ping-protect-versioning

ryanbas21:
# JIRA Ticket
N/A

## Description
This is a publish to fix the versioning problem presented when using `ping-protect`. We are now using the ~^so that any minor will be picked up.

https://pnpm.io/workspaces#publishing-workspace-packages

shows a reference to how this is translated

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,13 +7,12 @@
     }
   ],
   "commit": false,
-  "fixed": [["@forgerock/javascript-sdk"]],
+  "fixed": [["@forgerock/javascript-sdk"], ["@forgerock/ping-protect"]],
   "linked": [],
   "access": "public",
   "baseBranch": "master",
   "updateInternalDependencies": "patch",
   "ignore": [
-    "@forgerock/ping-protect",
     "@forgerock/token-vault",
     "autoscript-apps",
     "autoscript-suites",
--- a/packages/javascript-sdk/package.json
+++ b/packages/javascript-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@forgerock/javascript-sdk",
-  "version": "4.8.1",
+  "version": "4.8.2",
   "description": "ForgeRock JavaScript SDK",
   "author": "ForgeRock",
   "license": "MIT",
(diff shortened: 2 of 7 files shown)


packages/javascript-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@forgerock/javascript-sdk",
-  "version": "4.8.1",
+  "version": "4.8.2",
```
ancheetah, line 3: Should this be 4.8.3 or will the changeset auto bump this? The current release in Github is already 4.8.2

.changeset/config.json
```diff
@@ -7,13 +7,12 @@
     }
   ],
   "commit": false,
-  "fixed": [["@forgerock/javascript-sdk"]],
+  "fixed": [["@forgerock/javascript-sdk"], ["@forgerock/ping-protect"]],
```
ancheetah, line 10: This will cause protect to always be versioned along with the js sdk, changing the protect version from 4.6.0 --> 4.8.1 in this release. Is that correct/intended?

packages/ping-protect/package.json
```diff
@@ -20,6 +20,6 @@
   },
   "types": "./dist/index.ts.d.ts",
   "dependencies": {
-    "@forgerock/javascript-sdk": "workspace:*"
+    "@forgerock/javascript-sdk": "workspace:^"
```
ancheetah, line 23: Should there be a version number after this? What does the caret mean when it's alone like this?

.changeset/config.json
ryanbas21: It won't version it with the SDK, but the SDK would get bumped because it's a dependent.

ryanbas21: They would have to be grouped in the same sub-array to be fixed together (pretty sure but we can see in the release pr)

packages/javascript-sdk/package.json
ryanbas21: it would get bumped, i just manually made this up to date because i had to do a manual publish for the last release

ancheetah (approved):

PR #549, opened 2025-09-25
fix(package): remove invalid ./src/ export

buenoC6:
## Description

The "./src/" entry in the exports field was invalid since folder exports must point to a directory (ending with "/"), not to file patterns.  This entry is also redundant with "./src/*" and caused errors when parsing package.json in strict environments (e.g. Node 16+).

Added a patch changeset.

### DIFF
--- a/.changeset/fix-src-package.md
+++ b/.changeset/fix-src-package.md
@@ -1,0 +1,5 @@
+---
+"@forgerock/javascript-sdk": patch
+---
+
+Remove invalid `./src/` export from package.json to fix parsing errors in strict Node environments.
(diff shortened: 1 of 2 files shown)


ryanbas21:
> ## Description
> The "./src/" entry in the exports field was invalid since folder exports must point to a directory (ending with "/"), not to file patterns. This entry is also redundant with "./src/*" and caused errors when parsing package.json in strict environments (e.g. Node 16+).
> 
> Added a patch changeset.

Hey, thanks for this, and good catch. sorry i missed this for so long. I'll talk to product about a patch release for this in the future, i can't promise when appreciate the PR.

I may need to opent his myself since we don't use pull_request_target on ci for security and we need a key to run the commands in ci

ryanbas21:
re-opened this on another pr

PR #553, opened 2025-10-27
chore: fix publish workflow for real

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4448

## Description

- add `shell` to steps with `run`
- pass in env vars to sub action
- move beta slack notifications to new webhook

### DIFF
--- a/.github/actions/setup-publish/action.yml
+++ b/.github/actions/setup-publish/action.yml
@@ -1,6 +1,11 @@
 name: Setup publish
 description: Setup steps for publishing packages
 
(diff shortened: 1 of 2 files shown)


ryanbas21 (approved):

ancheetah:
Successfully tested manual trigger without publishing to npm. Will try an actual beta publish after this is merged in.

Note: Playwright artifacts don't seem to be picked up but I think we have the same problem in ping-javascript-sdk

PR #571, opened 2026-01-06
feat: add auto complete ui for AM-33773

cameronwhitworthforgerock:
# JIRA Ticket

[Ticket](https://pingidentity.atlassian.net/issues?jql=textfields%20~%20%22conditional%20UI%2A%22%20AND%20project%20%3D%20AME&selectedIssue=AME-33773)

## Description
Change enables support for autofill conditional UI in Javascript SDK.

Relates to https://pingidentity.atlassian.net/browse/OPENAM-23288

### DIFF
--- a/packages/javascript-sdk/src/fr-webauthn/fr-webauthn.mock.data.ts
+++ b/packages/javascript-sdk/src/fr-webauthn/fr-webauthn.mock.data.ts
@@ -489,3 +489,40 @@ export const webAuthnRegMetaCallbackJsonResponse = {
     },
   ],
 };
+
+export const webAuthnAuthConditionalMetaCallback = {
+  authId: 'test-auth-id-conditional',
+  callbacks: [
+    {
+      type: CallbackType.MetadataCallback,
+      output: [
+        {
+          name: 'data',
+          value: {
+            _action: 'webauthn_authentication',
+            challenge: 'JEisuqkVMhI490jM0/iEgrRz+j94OoGc7gdY4gYicSk=',
+            allowCredentials: '',
+            _allowCredentials: [],
+            timeout: 60000,
+            userVerification: 'preferred',
+            conditionalWebAuthn: true,
+            relyingPartyId: '',
+            _relyingPartyId: 'example.com',
+            extensions: {},
+            _type: 'WebAuthn',
+            supportsJsonResponse: true,
+          },
+        },
+      ],
+      _id: 0,
+    },
+    {
+      type: CallbackType.HiddenValueCallback,
+      output: [
+        { name: 'value', value: 'false' },
+        { name: 'id', value: 'webAuthnOutcome' },
+      ],
+      input: [{ name: 'IDToken1', value: 'webAuthnOutcome' }],
+    },
+  ],
+};
--- a/packages/javascript-sdk/src/fr-webauthn/index.ts
+++ b/packages/javascript-sdk/src/fr-webauthn/index.ts
@@ -13,6 +13,7 @@ import { CallbackType } from '../auth/enums';
 import type HiddenValueCallback from '../fr-auth/callbacks/hidden-value-callback';
 import type MetadataCallback from '../fr-auth/callbacks/metadata-callback';
 import type FRStep from '../fr-auth/fr-step';
+import { FRLogger } from '../util/logger';
 import { WebAuthnOutcome, WebAuthnOutcomeType, WebAuthnStepType } from './enums';
 import {
   arrayBufferToString,
@@ -30,6 +31,7 @@ import type {
 } from './interfaces';
 import type TextOutputCallback from '../fr-auth/callbacks/text-output-callback';
 import { parseWebAuthnAuthenticateText, parseWebAuthnRegisterText } from './script-parser';
+import { withTimeout } from '../util/timeout';
 
 // <clientdata>::<attestation>::<publickeyCredential>::<DeviceName>
 type OutcomeWithName<
@@ -44,6 +46,8 @@ type OutcomeWithName<
 type WebAuthnMetadata = WebAuthnAuthenticationMetadata | WebAuthnRegistrationMetadata;
 // Script-based WebAuthn
 type WebAuthnTextOutput = WebAuthnTextOutputRegistration;
+const TWO_SECOND = 2000;
+
 /**
  * Utility for integrating a web browser's WebAuthn API.
  *
@@ -58,6 +62,24 @@ type WebAuthnTextOutput = WebAuthnTextOutputRegistration;
  * } else if (stepType === WebAuthnStepType.Authentication) {
  *   // Authenticate with a registered device
  *   await FRWebAuthn.authenticate(step);
+ * }
+ * ```
+ *
+ * Conditional UI (Autofill) Support:
+ *
+ * ```js
+ * // Check if browser supports conditional UI
+ * const supportsConditionalUI = await FRWebAuthn.isConditionalUISupported();
+ *
+ * if (supportsConditionalUI) {
+ *   // The authenticate() method automatically handles conditional UI
+ *   // when the server indicates support via conditionalWebAuthn: true
+ *   // in the metadata. No additional code changes needed.
+ *   await FRWebAuthn.authenticate(step);
+ *
+ *   // For conditional UI to work in the browser, add autocomplete="webauthn"
+ *   // to your username input field:
+ *   // <input type="text" name="username" autocomplete="webauthn" />
  * }
  * ```
  */
@@ -95,7 +117,28 @@ abstract class FRWebAuthn {
   }
 
   /**
+   * Checks if the browser supports conditional UI (autofill) for WebAuthn.
+   *
+   * @return Promise<boolean> indicating if conditional mediation is available
+   */
+  public static async isConditionalUISupported(): Promise<boolean> {
+    if (!window.PublicKeyCredential) {
+      return false;
+    }
+
+    // Check if the browser supports conditional mediation
+    try {
+      return withTimeout(PublicKeyCredential.isConditionalMediationAvailable(), TWO_SECOND);
+    } catch {
+      FRLogger.warn('Conditional mediation check timed out');
+    }
+
+    return false;
+  }
+
+  /**
    * Populates the step with the necessary authentication outcome.
+   * Automatically handles conditional UI if indicated by the server metadata.
    *
    * @param step The step that contains WebAuthn authentication data
    * @return The populated step
@@ -108,12 +151,18 @@ abstract class FRWebAuthn {
 
       try {
         let publicKey: PublicKeyCredentialRequestOptions;
+        let useConditionalUI = false;
+
         if (metadataCallback) {
           const meta = metadataCallback.getOutputValue('data') as WebAuthnAuthenticationMetadata;
+
+          // Check if server indicates conditional UI should be used
+          useConditionalUI = meta.conditional === 'true';
           publicKey = this.createAuthenticationPublicKey(meta);
 
           credential = await this.getAuthenticationCredential(
             publicKey as PublicKeyCredentialRequestOptions,
+            useConditionalUI,
           );
           outcome = this.getAuthenticationOutcome(credential);
         } else if (textOutputCallback) {
@@ -121,6 +170,7 @@ abstract class FRWebAuthn {
 
           credential = await this.getAuthenticationCredential(
             publicKey as PublicKeyCredentialRequestOptions,
+            false, // Script-based callbacks don't support conditional UI
           );
           outcome = this.getAuthenticationOutcome(credential);
         } else {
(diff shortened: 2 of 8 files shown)


packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -300,18 +345,36 @@ abstract class FRWebAuthn {
...
+    };
+
+    // Add conditional mediation if requested and supported
+    if (useConditionalUI) {
+      const isConditionalSupported = await this.isConditionalUISupported();
+      if (isConditionalSupported) {
+        credentialRequestOptions.mediation = 'conditional' as CredentialMediationRequirement;
+      } else {
+        // eslint-disable-next-line no-console
+        console.warn('Conditional UI was requested, but is not supported by this browser.');
```
ryanbas21: we need to use `logger` for this, in case a user does not want console log's to happen. the logger module will call console logs if enabled.

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
```diff
@@ -77,12 +77,18 @@ interface WebAuthnRegistrationMetadata {
 }
 
 interface WebAuthnAuthenticationMetadata {
+  _action?: string;
```
ryanbas21: Can we have a better type than a string for this? seems like we can have `'webauthn_authentication'`, I assume the alternative is `'webauthn_registration'`?

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
```diff
@@ -77,12 +77,18 @@ interface WebAuthnRegistrationMetadata {
...
   allowCredentials?: string;
+  _allowCredentials?: PublicKeyCredentialDescriptor[];
   challenge: string;
   relyingPartyId: string;
+  _relyingPartyId?: string;
   timeout: number;
   userVerification: UserVerificationType;
+  conditionalWebAuthn?: boolean;
+  extensions?: Record<string, unknown>;
+  _type?: string;
```
ryanbas21: What are the possible `_type`'s that we can get? Ideally we'd like to be more strict than `string` where possible. In the test data I see `WebAuthn` but what other options are there? 

What role does `_type` play in this data if any as it relates to the sdk?

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -94,8 +112,27 @@ abstract class FRWebAuthn {
...
+   * @return Promise<boolean> indicating if conditional mediation is available
+   */
+  public static async isConditionalUISupported(): Promise<boolean> {
+    if (!window.PublicKeyCredential) {
+      return false;
+    }
+
+    // Check if the browser supports conditional mediation
+    if (typeof PublicKeyCredential.isConditionalMediationAvailable === 'function') {
+      return await PublicKeyCredential.isConditionalMediationAvailable();
```
ryanbas21: This method looks like it can `throw` so we may need to handle the error in case it does.

```
[Exceptions](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/isConditionalMediationAvailable_static#exceptions)
The returned [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) may be rejected with the following values:


SecurityError [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException)
The RP domain is not valid.
```



Otherwise we may leave the auth state in a failure and unrecoverable error.

packages/javascript-sdk/src/fr-webauthn/fr-webauthn.mock.data.ts
```diff
@@ -489,3 +489,40 @@ export const webAuthnRegMetaCallbackJsonResponse = {
...
+    {
+      type: CallbackType.MetadataCallback,
+      output: [
+        {
+          name: 'data',
+          value: {
+            _action: 'webauthn_authentication',
+            challenge: 'JEisuqkVMhI490jM0/iEgrRz+j94OoGc7gdY4gYicSk=',
+            allowCredentials: '',
+            _allowCredentials: [],
```
ryanbas21, line 505: So AM now sends both of these values? we need to take precedence on the array I assume because it can list the possible passkeys?

Is this AM being backwards compatible with the older `allowCredentials`?

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -60,6 +60,24 @@ type WebAuthnTextOutput = WebAuthnTextOutputRegistration;
...
+ *
+ * if (supportsConditionalUI) {
+ *   // The authenticate() method automatically handles conditional UI
+ *   // when the server indicates support via conditionalWebAuthn: true
+ *   // in the metadata. No additional code changes needed.
+ *   await FRWebAuthn.authenticate(step);
+ *
+ *   // For conditional UI to work in the browser, add autocomplete="webauthn"
+ *   // to your username input field:
+ *   // <input type="text" name="username" autocomplete="webauthn" />
```
ryanbas21, line 82: What happens if this isn't provided and the promise is hanging?

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -300,18 +345,36 @@ abstract class FRWebAuthn {
    * Retrieves the credential from the browser Web Authentication API.
    *
    * @param options The public key options associated with the request
+   * @param useConditionalUI Whether to use conditional UI (autofill)
    * @return The credential
    */
   public static async getAuthenticationCredential(
     options: PublicKeyCredentialRequestOptions,
+    useConditionalUI = false,
```
ryanbas21, line 358: as per the hanging promise comment, we may need to use an abort signal here to allow developers to abort the promise if the passkey option isn't there?

I may be not following this correctly, so please correct me where I'm wrong, but if a developer is using autofill ui, but fails to add the html correctly, can't we end up in a hanging promise state?

ryanbas21 (changes requested):
We need to ensure that because this is a newer browser feature we are backwards compatible with browsers that may not support this feature.

So graceful degradation of conditional ui should occur where the browser doesn't break the user and we can at least continue the flow as needed.

Any details regarding this would be helpful because i'm fairly new to this feature

This is specific to authentication right? Is this testable with virtual authenticators? If we can pre-populate a user with a passkey in AM, i think we should be able to create an e2e for this? 

If not, can we add an e2e test that at least can be run manually (doesn't need actual assertions, but it would be helpful for us to document how to do this work in code)

Also, please add a changeset to this PR, it should be a new feature, with a description of the feature being added so we can release correctly.

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: I've got a fix in based on code I've seen elsewhere in the code base. Feel free to correct me if I've mssed up.

packages/javascript-sdk/src/fr-webauthn/fr-webauthn.mock.data.ts
KMForgeRock: They are effectively identical as AM supports both for as you mentioned backwards compatibility. If you feel like this answers your question, give me a should and I can resolve this thread.

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Updated.

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Done, shout if this is correct/incorrect please.

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
KMForgeRock: Change made to literal string WebAuthn
Note for Kian: This means that the passed _type must be exactly that. If for whatever reason we want to pass in anything, then change it back to string. If we want to change it to discrete values then change it to 'X|Y|Z'

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
KMForgeRock: Done, see below

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Kian note:
There is a current question in my fix if after timeout we shout throw an exception or just return false. Either way we should make it so the admin can specify a timeout time (have a default just incase).
Throwing errors seems to be what we do in the SDK

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Also abort signal may be the right path, we would have to figure out how to do it.

packages/javascript-sdk/src/fr-webauthn/fr-webauthn.mock.data.ts
KMForgeRock: Commend addressed

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Comment resolved.

Removed console.warn

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
KMForgeRock: Done

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -94,8 +116,29 @@ abstract class FRWebAuthn {
...
+  public static async isConditionalUISupported(): Promise<boolean> {
+    if (!window.PublicKeyCredential) {
+      return false;
+    }
+
+    // Check if the browser supports conditional mediation
+    try{
+      return withTimeout(PublicKeyCredential.isConditionalMediationAvailable(), ONE_SECOND)
+    } catch {
+      throw new Error('Error determining conditional mediation support');
```
cerebrl: Should we throw here, or just return false? If this feature isn't supported, we can just fall back to "normal" behavior, yeah?

packages/javascript-sdk/src/fr-webauthn/index.ts
cerebrl: Yes, we throw in cases where the continuation is not possible, but in this case, we can consider this progressive enhancement, yeah? In other words, "conditional UI" isn't necessary for the basic function of WebAuthn, so we don't need to throw.

packages/javascript-sdk/src/fr-webauthn/index.ts
cerebrl: FRLogger itself handles the console log, so you can remove the duplicate `console.warn` statement.

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
cerebrl: You mention that this change is done, but I still see `string` as the value for `_action`. Am I missing something?

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
cerebrl: I see that `_type` is still `string`. Did you forget to push these changes?

cerebrl (changes requested):

ryanbas21:
Also please add a changeset. @cameronwhitworthforgerock

packages/javascript-sdk/src/fr-webauthn/index.ts
ryanbas21: @cerebrl it's possible that we haven't really "started" webauthn though. This flow can be activated before a username or password has entered, so the fallback to "webauthn" is unlikely because usually you need to progress to that point.

Right?

packages/javascript-sdk/src/fr-webauthn/interfaces.ts
KMForgeRock: If you check the file changes rather than these comments, you should see the change. For reason beyond me, github has decided to move this comment to Outdated so it can only be observed here.
Hope that makes sense.

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Definitely feels like there is no "right" answer. I think we (Liskov) will discuss this as a team and then go forward with that, if that sounds good to oyu.

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -300,18 +351,36 @@ abstract class FRWebAuthn {
...
     }
-    const credential = await navigator.credentials.get({ publicKey: options });
+    // Build the credential request options
+    const credentialRequestOptions: CredentialRequestOptions = {
+      publicKey: options,
+    };
+
+    // Add conditional mediation if requested and supported
+    if (useConditionalUI) {
+      const isConditionalSupported = await this.isConditionalUISupported();
```
KMForgeRock, line 373: This won't hang as the timeout was added above.

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -300,18 +351,36 @@ abstract class FRWebAuthn {
...
+      if (isConditionalSupported) {
+        credentialRequestOptions.mediation = 'conditional' as CredentialMediationRequirement;
+      } else {
+        // eslint-disable-next-line no-console
+        FRLogger.warn('Conditional UI was requested, but is not supported by this browser.');
+      }
+    }
+
+    const credential = await navigator.credentials.get(credentialRequestOptions);
```
KMForgeRock, line 382: This is for under discussion about hanging

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Talking to @george-bafaloukas-forgerock it may be worth keeping the ticket open to test when all other matters have been resolved, then we can fully test and make sure nothing it broken.

packages/javascript-sdk/src/fr-webauthn/index.ts
KMForgeRock: Talking to @george-bafaloukas-forgerock this should already be handled by the browser as it is a variable in credentialRequestOptions

KMForgeRock: I've double checked and timeout is passed in.

ryanbas21 (approved):

cerebrl (approved):

PR #577, opened 2026-02-20
Webauthn autocomplete extra fixes

thomas-schofield-fr:
# JIRA Ticket

AME-34348 AME-34340 AME-34349

## Description
WebAuthn improvements
* Fix parsing of WebAuthn scripts when `asScript` is true
* Improve handling when conditional mediation is not supported
* Enable re-invocation of WebAuthn requests
* Enable modification of options passed to navigator.credentials.get()

### DIFF
--- a/packages/javascript-sdk/src/fr-webauthn/fr-webauthn.test.ts
+++ b/packages/javascript-sdk/src/fr-webauthn/fr-webauthn.test.ts
@@ -131,7 +131,7 @@ describe('Test FRWebAuthn class with Conditional UI', () => {
 
   it('should detect if conditional UI is supported', async () => {
     vi.spyOn(window.PublicKeyCredential, 'isConditionalMediationAvailable').mockResolvedValue(true);
-    const isSupported = await FRWebAuthn.isConditionalUISupported();
+    const isSupported = await FRWebAuthn.isConditionalMediationSupported();
     expect(isSupported).toBe(true);
   });
 
--- a/packages/javascript-sdk/src/fr-webauthn/fr-webauthn.test.ts
+++ b/packages/javascript-sdk/src/fr-webauthn/fr-webauthn.test.ts
@@ -149,7 +149,7 @@ describe('Test FRWebAuthn class with Conditional UI', () => {
       _allowCredentials: [],
       timeout: 60000,
       userVerification: 'preferred',
-      conditionalWebAuthn: true,
+      mediation: 'conditional',
       relyingPartyId: '',
       _relyingPartyId: 'example.com',
       extensions: {},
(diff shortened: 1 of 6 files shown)


packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -599,6 +608,25 @@ abstract class FRWebAuthn {
...
     };
   }
+
+  private static createAbortController() {
+    window.PingWebAuthnAbortController?.abort();
+
+    const abortController = new AbortController();
+    window.PingWebAuthnAbortController = abortController;
+    return abortController;
+  }
```
ryanbas21, line 630: Why do we need the abort controller as part of the window? I don't see us using it from the Window, but maybe it's needed elsewhere?

packages/javascript-sdk/src/fr-webauthn/fr-webauthn.test.ts
```diff
@@ -149,7 +149,7 @@ describe('Test FRWebAuthn class with Conditional UI', () => {
       _allowCredentials: [],
       timeout: 60000,
       userVerification: 'preferred',
-      conditionalWebAuthn: true,
+      mediation: 'conditional',
```
ryanbas21, line 152: Is this the only possible value for `mediation`?

packages/javascript-sdk/src/fr-webauthn/index.ts
ForgeRockEmma: The webauthn processing persists across page loads, meaning that we need to store it somewhere that can survive that.

This method will abort the previous controller, cancelling previous flows, before creating a new one, meaning that it doesn't need to be referenced outside of this method.

packages/javascript-sdk/src/fr-webauthn/fr-webauthn.test.ts
ForgeRockEmma: Currently with the backend the only values this could be "conditional" or not present. There are other valid values however, as per https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/get#mediation. Currently AM does not send any of these other values.

packages/javascript-sdk/src/fr-webauthn/index.ts
ryanbas21: Got it, okay. hmm we typically don't want to modify the window but I see where your coming from. Let me think on this.

cerebrl (approved):

ryanbas21 (approved):

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -599,6 +620,25 @@ abstract class FRWebAuthn {
       },
     };
   }
+
+  private static createAbortController() {
+    window.PingWebAuthnAbortController?.abort();
+
+    const abortController = new AbortController();
+    window.PingWebAuthnAbortController = abortController;
```
cerebrl, line 628: This pattern is not typical for the SDK, modifying globals, so we'll need to find a different way to a address this in the new SDK. Since this is the old SDK, and we will be deprecating it soon, we'll go ahead and approve this to be a bit more time/effort efficient.

ancheetah (approved):

packages/javascript-sdk/src/fr-webauthn/index.ts
ForgeRockEmma: Thanks for the review! That makes sense.

ForgeRockEmma:
This is now ready to be merged. Thanks again!

PR #579, opened 2026-02-26
fix(ci): fix slack payload parsing in release workflow

ryanbas21:
## Summary
- **Fixed Slack payload parsing** that broke the last release CI run ([#411](https://github.com/ForgeRock/forgerock-javascript-sdk/actions/runs/19873730773)). The formatted multi-line mrkdwn text was interpolated raw into the YAML payload, causing `slackapi/slack-github-action` to fail with "Invalid input! Failed to parse contents of the provided payload."
- **Build a proper JSON payload via `jq -n --arg`** which correctly escapes newlines, asterisks, and backticks in package names/versions.
- **Use environment variables** for changesets output instead of direct `${{ }}` interpolation in `run:` blocks to prevent command injection.
- **Add `continue-on-error: true`** to both Slack notification steps so a webhook failure never blocks npm publishing or code coverage.

## Context
On the Dec 2 release (merge of changeset-release/master), the "Send GitHub Action data to a Slack workflow" step failed, which halted the job and caused code coverage to be skipped entirely.

## Test plan
- [ ] Trigger a snapshot release via `workflow_dispatch` and verify the Slack notification sends (or gracefully fails without blocking)
- [ ] On the next production release, confirm the formatted packages payload is valid JSON and the Slack message renders correctly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       - run: pnpm exec nx affected -t build lint test docs e2e-ci
 
       - name: Publish previews to Stackblitz on PR
-        run: pnpm pkg-pr-new publish './packages/*' --packageManager=pnpm
+        run: pnpm pkg-pr-new publish './packages/*' --packageManager=pnpm  --pnpm
 
       - uses: codecov/codecov-action@v5
         with:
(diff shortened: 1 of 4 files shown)


ancheetah (approved):
Approved. Slack message looks good and I also tested installing the js package from Stackblitz with no errors. Thanks Ryan

PR #581, opened 2026-03-03
SDKS-4549: Update Protect Callback With New Signals SDK Config

SteinGabriel:
# JIRA Ticket
- [SDKS-4549](https://pingidentity.atlassian.net/browse/SDKS-4549) 

## Description
This PR updates Ping Protect initialization to support the new `signalsInitializationOptions` callback property from AM.

## What Changed
- `PingOneProtectInitializeCallback#getConfig()` now:
 - returns `signalsInitializationOptions` directly when present as an object
 - falls back to legacy config construction when missing/invalid
- `ping-protect` types widened to support pass-through init option shapes
 (`SignalsInitializationOptions`, `ProtectNodeInitializeConfig`)
- `PIProtect.start()` now resumes behavioral collection when
 `behavioralDataCollection` is `true` or `'true'`
- Added/updated unit tests for:
 - valid `signalsInitializationOptions` pass-through
 - invalid/missing fallback behavior
 - `'true'`/`'false'` behavioral collection handling in `ping-protect`

## How Has This Been Tested?
Tested with the e2e autoscript apps as well as added a test assertion for our automated e2e test suite.


packages/ping-protect/src/lib/ping-signals-sdk.js
cerebrl, line 1: Just want to confirm that this is the latest and correct Signals version from the Protect team. If we've confirmed this version with them, then I'm good.

packages/ping-protect/src/lib/ping-protect.ts
```diff
@@ -69,7 +76,7 @@ export abstract class PIProtect {
     }
     await window._pingOneSignals.init(options);
 
-    if (options.behavioralDataCollection === true) {
+    if (options.behavioralDataCollection === true || options.behavioralDataCollection === 'true') {
       window._pingOneSignals.resumeBehavioralData();
```
vatsalparikh: We should call resumeBehavioralData() method defined below instead of repeating the code block here because resumeBehavioralData() internally calls window._pingOneSignals.resumeBehavioralData();

`  public static resumeBehavioralData(): void {
    window._pingOneSignals.resumeBehavioralData();
  }`

SteinGabriel: Yeah, I think that makes sense and I personally don't have anything against it, so I went ahead and updated it. However, I feel like this pattern could be a deliberate design choice, so @cerebrl if that's the case, please let us know. Thanks!

cerebrl (approved):

packages/ping-protect/src/lib/ping-signals-sdk.js
SteinGabriel: Yeah, I grabbed this code from this link: https://apps.pingone.com/signals/web-sdk/latest/signals-sdk.js, which was provided by Itai this past Monday on Slack.

PR #582, opened 2026-03-08
fix: for issues with _allowCredentials and device registration

thomas-schofield-fr:
# JIRA Ticket
Fixes for original changes made by:

1. AME-33781
2. AME-33773

## Description
Fixes 2 bugs:

1. Check for `isWebAuthnSupported()` should have been inverted when registering a new device.
2. `_allowCredentials` array entries have an `id` property which needed to be a typed array for `navigator.credentials.get()`.

### DIFF
--- a/packages/javascript-sdk/src/fr-webauthn/index.ts
+++ b/packages/javascript-sdk/src/fr-webauthn/index.ts
@@ -469,7 +469,7 @@ abstract class FRWebAuthn {
     options: PublicKeyCredentialCreationOptions,
   ): Promise<PublicKeyCredential | null> {
     // Feature check before we attempt registering a device
-    if (this.isWebAuthnSupported()) {
+    if (!this.isWebAuthnSupported()) {
       const e = new Error('PublicKeyCredential not supported by this browser');
       e.name = WebAuthnOutcomeType.NotSupportedError;
       throw e;
@@ -534,7 +534,14 @@ abstract class FRWebAuthn {
     // Use the structured _allowCredentials if available, otherwise parse the string format
     let allowCredentialsValue: PublicKeyCredentialDescriptor[] | undefined;
     if (_allowCredentials && Array.isArray(_allowCredentials)) {
-      allowCredentialsValue = _allowCredentials;
+      // The incoming _allowCredentials entries have an `id` property of type `Array`, which is rejected by `navigator.credentials.get()`.
+      // Converting it to a TypedArray here to meet the spec (https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialRequestOptions#id).
+      allowCredentialsValue = _allowCredentials.map((cred) => {
+        return {
+          ...cred,
+          id: new Int8Array(cred.id as unknown as number[]),
+        };
+      });
     } else {
       allowCredentialsValue = parseCredentials(allowCredentials || acceptableCredentials || '');
     }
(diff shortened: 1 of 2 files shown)


packages/javascript-sdk/src/fr-webauthn/helpers.ts
```diff
@@ -83,25 +83,6 @@ function parsePubKeyArray(value: string | unknown[]): PublicKeyCredentialParamet
   return JSON.parse(`[${value}]`);
 }
 
-function parseAllowCredentialsArray(
```
thomas-schofield-fr, line 86: This was unused

packages/javascript-sdk/src/fr-webauthn/index.ts
```diff
@@ -534,7 +534,14 @@ abstract class FRWebAuthn {
     // Use the structured _allowCredentials if available, otherwise parse the string format
     let allowCredentialsValue: PublicKeyCredentialDescriptor[] | undefined;
     if (_allowCredentials && Array.isArray(_allowCredentials)) {
-      allowCredentialsValue = _allowCredentials;
+      // The incoming _allowCredentials entries have an `id` property of type `Array`, which is rejected by `navigator.credentials.get()`.
+      // Converting it to a TypedArray here to meet the spec (https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialRequestOptions#id).
```
ryanbas21, line 538: I don't understand this comment. The type `PublicKeyCredentialDescriptor`  shows this

```ts
interface PublicKeyCredentialDescriptor {
    id: BufferSource;
    transports?: AuthenticatorTransport[];
    type: PublicKeyCredentialType;
}
```

This type comes from `lib.dom.ts` https://www.dec112.at/docs/ng112-js/1.9.4/interfaces/_internal_.PublicKeyCredentialDescriptor.html

Are we saying this is wrong?

`navigator.credentials.get` accepts  `CredentialRequestOptions`

```ts
interface CredentialRequestOptions {
    mediation?: CredentialMediationRequirement;
    publicKey?: PublicKeyCredentialRequestOptions;
    signal?: AbortSignal;
}
```

Pulling your branch down, and just looking at `cred` it seems like this is already the correct type?

Image

I'm a bit confused by this change.

thomas-schofield-fr: I think TypeScript just doesn't correctly verify the incoming `metadataCallback.getOutputValue('data')` is actually the correct type all the way through - specifically, the `id` property is `Array` which doesn't conform to `BufferSource`.
Screenshot 2026-03-10 at 16 19 18

If we don't make the change, the error from `navigator.credentials.get` is fairly clear:
Screenshot 2026-03-10 at 16 24 00

packages/javascript-sdk/src/fr-webauthn/helpers.ts
thomas-schofield-fr: reverted

packages/javascript-sdk/src/fr-webauthn/index.ts
ryanbas21: Does this happen on the latest version of the SDK before these changes? I feel like this would be a pretty breaking bug?

thomas-schofield-fr: It does not, it's a new issue since https://github.com/ForgeRock/forgerock-javascript-sdk/commit/03135cf543e3f694d48e6b9e0b9116ccf42737d1#diff-112486585734aee9407beea04e071c53e7eddce56a465b7462455fcb0e3bb346R514

ryanbas21 (approved):
Ran this locally with a local publish + webauthn flow and it worked.

PR #583, opened 2026-03-17
feat: Add Pre-Commit Copyright Date Check & Fix Automation

SteinGabriel:
## Summary

   Automates Ping-owned copyright year updates so developers no longer need
   to hand-edit headers. A pre-commit hook fixes and re-stages any staged
   files with stale Ping copyright years before lint and build checks run.

   Adds a pre-commit hook that automatically keeps Ping-owned copyright
   year headers up-to-date in staged files, removing the need for manual
   year updates on every change.

   ## Changes

   | File | Change |
   |---|---|
   | `tools/copyright/sync-header-years.mjs` | New script — fix/check logic |
   | `tools/copyright/sync-header-years.test.mjs` | 9 unit tests (Node built-in runner) |
   | `.husky/pre-commit` | Prepend copyright sync step |
   | `package.json` | Add `copyright:sync` and `copyright:check` scripts |

   ## How it works

   1. On `git commit`, the pre-commit hook reads the list of staged files.
   2. For each file it finds a stale Ping copyright year, it rewrites the
      header in place and re-stages the file.
   3. If any staged file has a Ping copyright header with **no recognisable
      year at all**, the commit is aborted with an error listing the incorrect files.
   4. All other files (non-Ping headers, binary, unreadable) are skipped.

   **Supported formats updated automatically:**

  `Copyright 2020 Ping Identity.` → `Copyright 2020 - 2026 Ping Identity.`
  `Copyright 2020-2024 Ping Identity.` → `Copyright 2020 - 2026 Ping Identity.` 
  `Copyright 2020-2026 Ping Identity.` → `Copyright 2020 - 2026 Ping Identity.` 
  `© Copyright 2020 - 2024 PingIdentity.` → `© Copyright 2020 - 2026 Ping Identity.` 
  `Copyright (c) 2023 - 2024 Ping Identity Corporation.` → `Copyright (c) 2023 - 2026 Ping Identity Corporation.`

   ## Testing
```bash
  node --test tools/copyright/sync-header-years.test.mjs
```

   ## Known gaps / follow-ups
   - [ ] CI enforcement (`copyright:check` on PRs) deferred to a follow-up.

   ## Rollout notes
   - No CI step added — pre-commit only for now.
   - Removing the hook does not break anything; it is additive only.

### DIFF
--- a/tools/copyright/sync-header-years.test.mjs
+++ b/tools/copyright/sync-header-years.test.mjs
@@ -1,0 +1,70 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import { hasInvalidPingCopyrightHeader, updateCopyrightYears } from './sync-header-years.mjs';
+
+test('updates stale range end year and keeps start year', () => {
+  const input = '/* Copyright 2020-2026 Ping Identity. All Rights Reserved */';
+  const actual = updateCopyrightYears(input, 2026);
+  assert.equal(actual, '/* Copyright 2020-2026 Ping Identity. All Rights Reserved */');
+});
+
+test('updates stale single year to a range preserving start year', () => {
+  const input = '/* Copyright 2025 - 2026 Ping Identity. All Rights Reserved */';
+  const actual = updateCopyrightYears(input, 2026);
+  assert.equal(actual, '/* Copyright 2025 - 2026 Ping Identity. All Rights Reserved */');
+});
+
+test('supports © and &copy; variants', () => {
+  const input = [
+    '/* © Copyright 2020-2026 Ping Identity. */',
+    '<!-- &copy; Copyright 2020-2026 Ping Identity. -->',
+  ].join('\n');
+  const actual = updateCopyrightYears(input, 2026);
+  assert.equal(
+    actual,
+    [
+      '/* © Copyright 2020-2026 Ping Identity. */',
+      '<!-- &copy; Copyright 2020-2026 Ping Identity. -->',
+    ].join('\n'),
+  );
+});
+
+test('does not update non-Ping headers', () => {
+  const input = '/* Copyright 2020-2025 Example Corp. */';
+  const actual = updateCopyrightYears(input, 2026);
+  assert.equal(actual, input);
+});
+
+test('updates Ping Identity Corporation ranges with spaces and (c)', () => {
+  const input = '/* Copyright (c) 2023 - 2026 Ping Identity Corporation. All right reserved. */';
+  const actual = updateCopyrightYears(input, 2026);
+  assert.equal(
+    actual,
+    '/* Copyright (c) 2023 - 2026 Ping Identity Corporation. All right reserved. */',
+  );
+});
+
+test('updates Ping Identity Corporation stale single year with (c) to range', () => {
+  const input = '/* Copyright (c) 2025 - 2026 Ping Identity Corporation. All right reserved. */';
+  const actual = updateCopyrightYears(input, 2026);
+  assert.equal(
+    actual,
+    '/* Copyright (c) 2025 - 2026 Ping Identity Corporation. All right reserved. */',
+  );
+});
+
+test('flags Ping headers without a valid year', () => {
+  const input = '/* Copyright Ping Identity Corporation. All right reserved. */';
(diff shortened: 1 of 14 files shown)


tools/copyright/sync-header-years.test.mjs
```diff
@@ -0,0 +1,82 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import { hasInvalidPingCopyrightHeader, updateCopyrightYears } from './sync-header-years.mjs';
+
+test('updates stale range end year and keeps start year', () => {
+  const input = '/* Copyright 2020 - 2026 Ping Identity. All Rights Reserved */';
+  const actual = updateCopyrightYears(input, 2026);
+  assert.equal(actual, '/* Copyright 2020 - 2026 Ping Identity. All Rights Reserved */');
```
ancheetah: In a lot of these tests the `input` looks identical to the value being asserted. For example, in this test shouldn't the input have something like years `2020 - 2025` before it is updated?

ancheetah (commented):
This is amazing. Thank you for doing this Gabriel! Will save us a lot of time. Just left a comment on the tests.

tools/copyright/sync-header-years.test.mjs
SteinGabriel: I could swear I had this right and I was loosing my mind wondering why my input date updates weren't showing up, until I realized that the pre-commit script was actually "fixing" the input date header in the test files, making it the same as the value being asserted. lol
Anyways, I added a check to prevent it from running on specific paths, such as test files. Thanks for pointing it out!

ryanbas21 (commented):
This is cool. I'd like us to weigh this approach versus https://www.npmjs.com/package/eslint-plugin-license-header as an eslint rule with auto fixing.

Could you, when available to, maybe present the differences / overlap / benefits to each approach? Maybe theres a better eslint rule out there also, I just happened to find this one and it seems up to date / maintained.

SteinGabriel:
> This is cool. I'd like us to weigh this approach versus [npmjs.com/package/eslint-plugin-license-header](https://www.npmjs.com/package/eslint-plugin-license-header) as an eslint rule with auto fixing.
> 
> Could you, when available to, maybe present the differences / overlap / benefits to each approach? Maybe theres a better eslint rule out there also, I just happened to find this one and it seems up to date / maintained.

The main limitation I can see is that ESLint rules enforce a fixed template. They are great at spotting missing headers, but they cannot keep the start year and only bump the end year. If we used a standard rule, I think we would lose our start years (e.g., 2020–2025 would become 2026). The pre-commit script is much more flexible for those range updates. Also, we have some pattern variation in the headers (such as `Copyright (c)`, `© Copyright`, and `&copy; Copyright`) which an ESLint rule might not handle.

That being said, ESLint rules could add value in enforcing header presence on new files. We could have an ESLint rule scoped to detecting missing headers only (with auto-fix disabled for year ranges), while keeping this pre-commit script for year updates.

ryanbas21:
> (e.g., 2020–2025 would become 2026).

I think this is actually okay. I'm not 100% sure if the year range is a hard requirement. I think adding both could be valuable though, even if we don't allow the eslint rule to autofix.

SteinGabriel:
> > (e.g., 2020–2025 would become 2026).
> 
> I think this is actually okay. I'm not 100% sure if the year range is a hard requirement. I think adding both could be valuable though, even if we don't allow the eslint rule to autofix.

Yeah, I'm also not sure if it’s a hard requirement, but I’ve seen that pattern in every repo so far. Sounds good though, I'll add the eslint rule. Thanks!

SteinGabriel:
@ryanbas21 I just pushed a change to enforce copyright headers using [eslint-plugin-header](https://www.npmjs.com/package/eslint-plugin-header). It works well, but there's a small quirk: since our pre-commit lint hook uses `--fix`, it’ll automatically add a template into any file missing a header. It won't block the new commit to be created, you’ll just see the modified file in your local changes afterward so you can fix the year by replacing the template's placeholder `<current_year>` or simply discard it. 

I believe it might make sense if we want to enforce that source files should have a copyright header, and it's also not a blocker when creating the commit, but I'm not really sure how annoying this might be. 
Let me know what you think. Thanks!

PR #584, opened 2026-03-24
fix: prevent destructuring error when reading from stored values

CEOehis:
## Description

`getStoredAuthUrlValues` can sometimes return null (e.g when the item is not found in storage) and attempting to destructure the return value causes an uncaught runtime error.

Changes:
- default to empty object with nullish coallescing before destructuring
- update `getStoredAuthUrlValues` type definition to reflect actual return type

This ensures the proper error handling downstream, where `state` and `verifier` are checked for undefined values, works as intended rather than crashing on destructuring.


ryanbas21:
@CEOehis thanks, the team will look to review this soon.

cerebrl (approved):
This looks good to me. I'll leave it to @ryanbas21 to review and merge when appropriate.

PR #587, opened 2026-05-04
fix: retrieve WebAuthn credentials inside try/catch OPENAM-26284

ForgeRockEmma:
# JIRA Ticket

OPENAM-26284 This fixes an issue where errors were not being caught and reported to AM

## Description

In a previous restructure, getAuthenticationCredential was moved outside of the catch block which populates the HiddenValueCallback. This meant that errors would not be reported to AM.


cerebrl (approved):
This looks good. I'd like @ryanbas21 to take a look as well.

ryanbas21 (approved):

